PR #7, opened 2017-09-29
Fix readtimeout error not throwing

atrakh:


src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -92,7 +92,7 @@ public EventSourceService(Configuration configuration)
                     {
                         using (var reader = new StreamReader(stream))
                         {
-                            while (!cancellationToken.IsCancellationRequested && !reader.EndOfStream)
+                            while (!cancellationToken.IsCancellationRequested)
```
arun251: Why don't we want `!reader.EndOfStream`?

atrakh: Strangely, accessing `reader.EndOfStream` causes `readTimeoutTask` to never resolve. Even a simple logging statement which accesses `reader.EndOfStream` reproduces this behavior.

atrakh:
Tests were failing because they rely on EventSource to close when the end of the stream was reached. This is not the functionality we want for our SDK. So, adding a configurable `closeOnEndOfStream` property allows for tests to pass, and for read timeout to work.

Re-review?

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -107,13 +107,22 @@ public EventSourceService(Configuration configuration)
...
+                                }
                             }
                         }
                     }
 
                     OnConnectionClosed();
                 }
             }
+            catch (HttpRequestException e) {
+                OnConnectionClosed();
```
ryanfollmer: You shouldn't need to call the OnConnectionClosed method here. The calling code handles basic Exception types and will call Connection Close - see the catch block in EventSource.cs in the ConnectToEventSourceAsync method: `CloseAndRaiseError(e);`

atrakh: I believe `OnConnectionClosed` should be the method passed into EventSource @ initialization time. EventSource.cs does close the client on error, but does not call `OnConnectionClosed`

ryanfollmer: EventSource.cs doesn't need to call OnConnectionClosed, rather the EventSourceService.cs calls this on the EventSource.cs. In the event of an error, EventSource itself will set the Ready State to Closed.

PR #8, opened 2017-10-05
Removed closeOnEndOfStream and resolve Unit Test issues with stream

ryanfollmer:
1. Removed the closeOnEndOfStream configuration property.
2. Modified the EventSourceTests.cs class to resolve the issue of the tests not completing.
3. Refactored the Event Source Service class to allow for unit testing the read timeout by way of a test double.
4. Added 2 Unit tests (positive and negative) to verify the read timeout functionality:
   a. When_response_exceeds_read_timeout_then_read_timeout_exception_occurs
   b. When_response_does_not_exceed_read_timeout_then_expected_message_event_occurs
5. Added another Console project targeting .NET Core: EventSource-ConsoleApp-DotNetCore.csproj


console-app/Program.cs
```diff
@@ -15,21 +15,21 @@ static void Main(string[] args)
         {
 
             Dictionary<string, string> headers =
-                new Dictionary<string, string> {{"Authorization", "<Insert Auth Key>"}};
+                new Dictionary<string, string> {{"Authorization", "<Insert Auth Key>" } };
```
atrakh: recommend adding spacing on opening bracket

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -128,9 +121,11 @@ public TimeSpan MaximumDelayRetryDuration
...
         /// <param name="logger">The logger used for logging internal messages.</param>
-        /// <param name="closeOnEndOfStream>If true, the EventSource connection will be closed when the end of the stream is reached.</param>
-        /// <exception cref="ArgumentOutOfRangeException">If the delayRetryDuration value is greater than 30 seconds, an ArgumentOutOfRangeException will be thrown.</exception>
-        public Configuration(Uri uri, HttpMessageHandler messageHandler = null, TimeSpan? connectionTimeOut = null, TimeSpan? delayRetryDuration = null, TimeSpan? readTimeout = null, IDictionary<string, string> requestHeaders = null, string lastEventId = null, ILogger logger = null, bool closeOnEndOfStream = false)
+        /// <exception cref="ArgumentNullException">Throws ArgumentNullException if the uri parameter is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Throws ArgumentOutOfRangeException if the connectionTimeOut value is less than zero.</exception>
+        /// <exception cref="ArgumentOutOfRangeException">Throws ArgumentOutOfRangeException if the delayRetryDuration value is greater than 30 seconds.</exception>
```
arun251: It might be better to combine  all the cases for a single exception into a single entry: https://stackoverflow.com/questions/6434499/xml-comments-how-to-comment-multiple-reasons-for-an-exception

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -356,20 +371,91 @@ public void Exponential_backoff_should_not_exceed_maximum_in_test_loop()
...
+
+        [Fact]
+        public async Task When_response_exceeds_read_timeout_then_read_timeout_exception_occurs()
+        {
+            // Arrange
+            var commentSent = ":";
+
+            var handler = new StubMessageHandler();
+            handler.QueueStringResponse(commentSent);
+            //handler.QueueResponse(new HttpResponseMessage(HttpStatusCode.NoContent));
```
arun251: Is this line obsolete?

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -356,20 +371,91 @@ public void Exponential_backoff_should_not_exceed_maximum_in_test_loop()
...
             //// Assert
             Assert.NotEmpty(backoffs);
             Assert.True(backoffs.Distinct().Count() == backoffs.Count());
         }
+
+
+        [Fact]
+        public async Task When_response_exceeds_read_timeout_then_read_timeout_exception_occurs()
+        {
+            // Arrange
```
arun251: You could add a bit more to the comment or just remove it. ;)

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -140,6 +106,36 @@ public EventSourceService(Configuration configuration)
...
+
+        protected virtual async Task ProcessResponseFromReaderAsync(Action<string> processResponse, IStreamReader reader,
+            CancellationToken cancellationToken)
+        {
+
+            // In a loop, reading from a stream. while reading from the stream, process the lines in the stream.
+            // Reset the timer after processing each line (reading from the stream).
+            // If the read time out occurs, throw exception and exit method.
+
+            while (!cancellationToken.IsCancellationRequested)
```
arun251: Do we not need a check against `reader.EndOfStream` after all?

console-app/Program.cs
ryanfollmer: Will do

src/LaunchDarkly.EventSource/Configuration.cs
ryanfollmer: Yep, makes sense. I'll update this.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
ryanfollmer: Yeah, I'll remove it

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
ryanfollmer: I'll remove the comment.

src/LaunchDarkly.EventSource/EventSourceService.cs
ryanfollmer: While reviewing the .net framework code for the EndOfStream property, it turns out the this property actually tries to read from the internal stream. What this does when connected to the EventSource API is wait for a "read" (response) from the stream and never proceeds into the while loop. So it turns out this does cause an issue because the read timeout is inside the while loop (which never gets reached in the case where the "read" is waiting). 

I tried moving the read timer outside this while loop but that resulted in a different problem in that the timer wasn't reset after parsing each "read" in the stream. 

I believe the code is correct now. Read timeouts do occur when they should and the timer is reset after each read.

PR #11, opened 2017-10-31
[CH7279] Inline ES connection retries

### LINKED ISSUE #9: Remove dependency on Polly-Signed
The dependency on `Polly-Signed` has been problematic in our project, which uses `Polly` (the not-signed version). We've had to change from `Polly` to `Polly-Signed` in each of our projects, which is a bit of a hassle. 

It looks like there's just one retry policy used here; can I suggest that the retry logic be written manually and the dependency on `Polly-Signed` removed? That way this library would have less of an impact on client projects.

atrakh:
This resolves #9

### DIFF
--- a/src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
+++ b/src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
@@ -4,13 +4,13 @@ using System.Text;
 
 namespace LaunchDarkly.EventSource
 {
-    internal class ExponentialBackoffWithDecorrelation
+    public class ExponentialBackoffWithDecorrelation
     {
-        private readonly double _minimumDelay;
-        private readonly double _maximumDelay;
+        private readonly TimeSpan _minimumDelay;
+        private readonly TimeSpan _maximumDelay;
         private readonly Random _jitterer = new Random();
 
-        public ExponentialBackoffWithDecorrelation(double minimumDelay, double maximumDelay)
+        public ExponentialBackoffWithDecorrelation(TimeSpan minimumDelay, TimeSpan maximumDelay)
         {
             _minimumDelay = minimumDelay;
             _maximumDelay = maximumDelay;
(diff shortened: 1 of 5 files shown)


src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
```diff
@@ -8,29 +8,19 @@ internal class ExponentialBackoffWithDecorrelation
...
-            nextDelay = Math.Min(_maximumDelay, Math.Max(_minimumDelay, _currentDelay * 3 * nextJitter));
-
-            if (nextDelay == _currentDelay)
-                nextDelay = GetBackOff().TotalMilliseconds;
-
-            _currentDelay = nextDelay;
-
-            return TimeSpan.FromMilliseconds(_currentDelay);
+            double nextDelay = Math.Min(_maximumDelay, _minimumDelay * Math.Pow(2, reconnectAttempts));
+            nextDelay = nextDelay / 2 + _jitterer.Next((int)nextDelay) / 2;
```
arun251: I don't think we need to use `double` at all. If we can get away with `int` for milliseconds everywhere, that would be fine.

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
```diff
@@ -4,35 +4,32 @@
...
+            int nextDelay = Convert.ToInt32(Math.Min(_maximumDelay.TotalMilliseconds, _minimumDelay.TotalMilliseconds * Math.Pow(2, _reconnectAttempts++)));
+            nextDelay = nextDelay / 2 + _jitterer.Next(nextDelay) / 2;
+            return TimeSpan.FromMilliseconds(nextDelay);
         }
 
+        public int GetReconnectAttemptCount() {
+            return _reconnectAttempts;
+        }
 
+        public void IncrementReconnectAttemptCount() {
```
jkodumal: We probably need to clear this on successful connect. Otherwise an error in the past could be impacting current reconnection times.

atrakh: ✔️

jkodumal: magic!

PR #14, opened 2017-12-04
make HTTP status code available to error handlers

eplusminus:
This is necessary in order to implement the desired "stop trying to do things if we get a 401 error" logic in the client. Previously, the information we passed to EventSource error handlers only contained a string description of the error; this adds an exception subclass that contains the status code.

### DIFF
--- a/test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
+++ b/test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
@@ -336,7 +336,8 @@ namespace LaunchDarkly.EventSource.Tests
             //// Assert
             Assert.NotNull(raisedEvent);
             Assert.Equal(evt, raisedEvent.Sender);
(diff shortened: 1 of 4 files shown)


test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -469,5 +472,23 @@ public void Exponential_backoff_should_reset_when_reconnect_count_resets()
             Assert.True(wasMessageReceivedEventRaised);
 
         }
+
+        [Fact]
+        public async Task When_error_handler_closes_event_source_no_reconnect_attempt_is_made()
```
eplusminus: This test is unrelated to the error reporting changes, but verifies another piece of behavior that's important for the upcoming client changes: if an error handler calls `Close` (for instance because it saw a 401 error), we want to make sure the EventSource respects this and doesn't try to reconnect.

eplusminus:
Just realized I also need to make those exception classes public if the caller is going to reference them.

PR #16, opened 2017-12-17
Switching to call the correct logging extension methods

JeffAshton:


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -125,8 +125,8 @@ public EventSource(Configuration configuration)
                 }
                 catch (Exception e)
                 {
-                    _logger.LogError("Encountered an error connecting to EventSource: {0}", e.Message);
-                    _logger.LogDebug(e.ToString());
+                    _logger.LogError( 0, e, "Encountered an error connecting to EventSource: {0}", e.Message);
```
JeffAshton: The exception is really important to pass through.  It contains the details of what actually happened.

When you're using async/await, and an AggregateException is thrown, you end up with this log message:

```Encountered an error connecting to EventSource: One or more errors occurred.```

By not passing along the exception, we have no way to know exactly what's going on.

File: src/LaunchDarkly.EventSource/EventSource.cs [hunk 2]
```diff
@@ -125,8 +125,8 @@ public EventSource(Configuration configuration)
...
+                    _logger.LogDebug( "{0}", e);
```
JeffAshton: As mentioned above, the exception should just be passed into the LogError method, but if you are going to log the raw error details this way, you only want to call ToString() if the Debug log level is enabled.

By switching this to a format string, you're guaranteed

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -135,7 +135,7 @@ public EventSource(Configuration configuration)
             if (_backOff.GetReconnectAttemptCount() > 0 && _retryDelay > TimeSpan.FromMilliseconds(0))
             {
                 TimeSpan sleepTime = _backOff.GetNextBackOff();
-                _logger.LogInformation("Waiting " + sleepTime.TotalMilliseconds + " milliseconds before reconnecting...");
+                _logger.LogInformation("Waiting {0} milliseconds before reconnecting...", sleepTime.TotalMilliseconds);
```
JeffAshton: Similar to above, you don't want to do the string concatenation unless the Information log level is actually enabled.

atrakh:
Hi @JeffAshton,

We'll be adding this in to the next .NET SDK release along with your suggestions in https://github.com/launchdarkly/.net-client/pull/57.

Thanks!

PR #28, opened 2018-06-01
Add HTTP method/body params to EventSource config

eli-darkly:
We'll need to be able to send a REPORT body for mobile/Xamarin.

### DIFF
--- a/src/LaunchDarkly.EventSource/Configuration.cs
+++ b/src/LaunchDarkly.EventSource/Configuration.cs
@@ -11,6 +11,11 @@ namespace LaunchDarkly.EventSource
     /// </summary>
     public sealed class Configuration
     {
+        #region Types
+
+        public delegate HttpContent HttpContentFactory();
+
+        #endregion
 
         #region Private Fields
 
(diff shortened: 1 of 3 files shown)


src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -129,7 +149,8 @@ public TimeSpan MaximumDelayRetryDuration
         ///     <p>- or - </p>
         ///     <p><paramref name="readTimeout"/> is less than zero. </p>
         /// </exception>
-        public Configuration(Uri uri, HttpMessageHandler messageHandler = null, TimeSpan? connectionTimeOut = null, TimeSpan? delayRetryDuration = null, TimeSpan? readTimeout = null, IDictionary<string, string> requestHeaders = null, string lastEventId = null, ILog logger = null)
+        public Configuration(Uri uri, HttpMessageHandler messageHandler = null, TimeSpan? connectionTimeOut = null, TimeSpan? delayRetryDuration = null, TimeSpan? readTimeout = null, IDictionary<string, string> requestHeaders = null, string lastEventId = null, ILog logger = null,
```
arun251: I had to remind myself that C# has a language feature where constructor parameters with defaults do not need to be specified, which is why we don't have a builder here.

eli-darkly: Yeah, I'm not crazy about how this looks either. I would rather have a builder. But then our fork would be diverging quite a bit for reasons unrelated to functionality.

PR #29, opened 2018-06-01
add a configuration builder for EventSource

eli-darkly:


src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -59,7 +62,10 @@ public sealed class Configuration
         /// <value>
         /// The <see cref="TimeSpan"/> before reading times out. The default value is 300,000 milliseconds (5 minutes).
         /// </value>
-        public TimeSpan ReadTimeOut { get; }
```
eli-darkly: I deprecated the `...TimeOut` names because the .NET framework uses `Timeout` (one word).

PR #33, opened 2018-08-02
when there's a backoff delay, actually do the delay

eli-darkly:
https://app.clubhouse.io/launchdarkly/story/21565/net-sdk-does-not-wait-before-reconnecting-to-stream

I tested this by running the test console app in this repo against a local streamer, then killing the streamer. Prior to this fix, it would log messages of "reconnecting after ____ milliseconds" but the next reconnect attempt would happen immediately.

### DIFF
--- a/console-app-dotnetcore/Program.cs
+++ b/console-app-dotnetcore/Program.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Threading;
 using LaunchDarkly.EventSource;
+using Common.Logging;
+using Common.Logging.Simple;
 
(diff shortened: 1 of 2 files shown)


console-app-dotnetcore/Program.cs
```diff
@@ -10,6 +12,8 @@ class Program
 
         static void Main(string[] args)
         {
+            LogManager.Adapter = new Common.Logging.Simple.DebugLoggerFactoryAdapter();
```
eli-darkly: Just making it so you can see the log output of the console app - at least, if you're running in the IDE (it will appear in the Application Output panel).

PR #36, opened 2018-08-28
fix stream reconnect logic

eli-darkly:
As reported [here](https://github.com/launchdarkly/dotnet-client/issues/88), the stream would not reconnect if it received an HTTP error code from the server. This was due to accidentally reusing the same CancellationToken to mean both "give up on the current HTTP request" and "give up permanently on the whole event source".

We did in fact have some unit tests with a fake HTTP server returning various errors. But the unit tests all relied on the wrong behavior, i.e. they expected the stream to close itself permanently instead of reconnecting. So I've reworked the tests too - I'd particularly appreciate comments on whether you think the tests now have adequate coverage of the various failure modes. (I also tested this manually against a fake stream.)

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSource.cs
+++ b/src/LaunchDarkly.EventSource/EventSource.cs
@@ -25,8 +25,9 @@ namespace LaunchDarkly.EventSource
         private string _eventName = Constants.MessageField;
         private string _lastEventId;
         private TimeSpan _retryDelay;
-        private CancellationTokenSource _pendingRequest;
         private readonly ExponentialBackoffWithDecorrelation _backOff;
+        private CancellationTokenSource _currentRequestToken;
+        private ReadyState _readyState;
 
         #endregion
 
@@ -65,8 +66,20 @@ namespace LaunchDarkly.EventSource
         /// </value>
         public ReadyState ReadyState
         {
-            get;
-            private set;
+            get
+            {
+                lock(this)
+                {
+                    return _readyState;
+                }
+            }
+            private set
+            {
+                lock(this)
+                {
+                    _readyState = value;
+                }
+            }
         }
 
         internal TimeSpan BackOffDelay
@@ -88,13 +101,11 @@ namespace LaunchDarkly.EventSource
         /// configuration</exception>
         public EventSource(Configuration configuration)
         {
-            ReadyState = ReadyState.Raw;
+            _readyState = ReadyState.Raw;
 
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
 
             _logger = _configuration.Logger ?? LogManager.GetLogger(typeof(EventSource));
-
-            _pendingRequest = new CancellationTokenSource();
 
             _retryDelay = _configuration.DelayRetryDuration;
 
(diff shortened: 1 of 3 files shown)


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -114,14 +125,22 @@ public EventSource(Configuration configuration)
         /// <exception cref="InvalidOperationException">The method was called after the connection <see cref="ReadyState"/> was Open or Connecting.</exception>
         public async Task StartAsync()
         {
-            var cancellationToken = _pendingRequest.Token;
-            while (!cancellationToken.IsCancellationRequested)
```
eli-darkly: This was the main problem.

`_pendingRequest.Token` was being used to keep track of the state of the current HTTP request - if we received bad data, we would cancel this token which would then terminate the request.

We do not want that to cause the whole stream to shut down. And we were apparently trying to avoid doing so: on [this line](https://github.com/launchdarkly/dotnet-eventsource/pull/36/files#diff-896d0f336a16ef11d2f9c3bce9244546L167), we were replacing `_pendingRequest` with a new CancellationTokenSource before cancelling the old one, in the case of a non-fatal error. But that didn't work, since the `while` loop here was still checking the _old_ `_pendingRequest`.

In any case, it's only the HTTP request that really needs a CancellationToken. For the while loop here, it makes more sense to just check explicitly whether we've been shut down.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -377,21 +374,19 @@ public async Task Given_204_when_the_http_response_is_received_then_error_event_
...
             var evt = new EventSource(new Configuration(_uri, handler));
 
+            var receiver = new ErrorReceiver(evt);
+
             //Act
-            var raisedEvent = await Assert.RaisesAsync<ExceptionEventArgs>(
-                h => evt.Error += h,
-                h => evt.Error -= h,
-                () => evt.StartAsync());
+            await evt.StartAsync();
```
eli-darkly: This is an example of the unit tests verifying the wrong behavior. `await evt.StartAsync()` will only return once the stream has been completely shut down. An HTTP error is not supposed to shut it down - it should reconnect. The test only worked because the stream was broken.

In the new version, `ErrorReceiver` takes care of closing the stream once it has received an error.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -114,14 +125,22 @@ public EventSource(Configuration configuration)
...
+            while (ReadyState != ReadyState.Shutdown)
             {
                 await MaybeWaitWithBackOff();
                 try
                 {
-                    await ConnectToEventSourceAsync(cancellationToken);
-                    _backOff.ResetReconnectAttemptCount();
```
eli-darkly: I've moved this call to `ResetReconnectAttemptCount` to [here](https://github.com/launchdarkly/dotnet-eventsource/pull/36/files#diff-896d0f336a16ef11d2f9c3bce9244546R215), to make it behave the way I think we always intended it to: it should reset the backoff count if it has successfully connected to the server. Putting it on line 124 was wrong because `await ConnectToEventSourceAsync` will only return when the connection has been closed.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -252,12 +276,19 @@ private void ProcessResponseContent(string content)
 
         private void SetReadyState(ReadyState state, Action<StateChangedEventArgs> action = null)
         {
-            if (ReadyState == state) return;
-
-            ReadyState = state;
+            lock (this)
+            {
+                if (_readyState == state || _readyState == ReadyState.Shutdown)
```
eli-darkly: The second part of this test is important in order to avoid a race condition where `Close()` gets called, setting the state to Shutdown, and _then_ an error handler fires and sets it to Closed. Once it is Shutdown, it should stay that way.

Issue #37, opened 2018-09-18
DelayRetryDuration not working as intended


magicxor:
### Steps to reproduce:

1) run the server-side php script which will close a connection every time after it sends the data:
```php
<?php
header('Content-Type: text/event-stream');
header('Cache-Control: no-cache'); // recommended to prevent caching of event data.

/**
 * Constructs the SSE data format and flushes that data to the client.
 *
 * @param string $id Timestamp/id of this connection.
 * @param string $msg Line of text that should be transmitted.
 */
function sendMsg($id, $msg) {
  echo "id: $id" . PHP_EOL;
  echo "data: $msg" . PHP_EOL;
  echo PHP_EOL;
  ob_flush();
  flush();
}

$serverTime = time();

sendMsg($serverTime, 'server time: ' . date("h:i:s", time()));
```

2) run the console app:
```csharp
        static async Task Main(string[] args)
        {
            var config = Configuration.Builder(new Uri("http://localhost:8000/your_script.php"))
                .ConnectionTimeout(TimeSpan.FromSeconds(1))
                .DelayRetryDuration(TimeSpan.FromSeconds(3))
                .ReadTimeout(TimeSpan.FromMinutes(4))
                .Build();

            var evt = new EventSource(config);

            evt.Opened += (sender, eventArgs) => { Console.WriteLine("Opened"); };
            evt.Error += (sender, eventArgs) => { Console.WriteLine("Error"); };
            evt.CommentReceived += (sender, eventArgs) => { Console.WriteLine("CommentReceived"); };
            evt.MessageReceived += (sender, eventArgs) => { Console.WriteLine($"MessageReceived: {eventArgs.Message.Data}"); };
            evt.Closed += (sender, eventArgs) => { Console.WriteLine("Closed"); };

            try
            {
                await evt.StartAsync();
            }
            catch (Exception ex)
            {
                Console.WriteLine("Closed");
            }

            Console.ReadLine();
        }
```

3) There should be a delay = 3 seconds between reconnects, but EventSource reconnects immediately. It looks like it ignores DelayRetryDuration(...) setting.

### My workaround is to change some code:

- in ExponentialBackoffWithDecorrelation.cs:
```csharp
        public TimeSpan GetNextBackOff(TimeSpan retryDelay)
        {
            _reconnectAttempts++;
            return retryDelay;
        }
```
- in EventSource.cs:
```csharp
        public async Task StartAsync()
        {
[...]

Issue #24, opened 2018-02-22
Null Reference Exceptions in EventSource


JeffAshton:
**Message:** Object reference not set to an instance of an object.
**Type:** System.NullReferenceException

```
   at LaunchDarkly.EventSource.EventSource.ProcessResponseContent(String content)
   at LaunchDarkly.EventSource.EventSourceService.<ProcessResponseFromReaderAsync>d__13.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LaunchDarkly.EventSource.EventSourceService.<ConnectToEventSourceApi>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LaunchDarkly.EventSource.EventSourceService.<GetDataAsync>d__10.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LaunchDarkly.EventSource.EventSource.<ConnectToEventSourceAsync>d__37.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LaunchDarkly.EventSource.EventSource.<StartAsync>d__32.MoveNext()
```
Occurs around the same time tasks are cancelled:

![image](https://user-images.githubusercontent.com/12101120/36539283-4a506da2-17a4-11e8-9f00-3cf7d171a57f.png)

eli-darkly:
I'm awfully sorry we didn't follow up on this issue sooner - it seems to have completely fallen through the cracks somehow. We should have a fix pretty soon. I wanted to confirm, though: am I right that the exception did not propagate to application code, but was caught and logged inside the EventSource?

[...]

PR #38, opened 2018-09-25
add null guard for stream reader output

eli-darkly:
The lack of a null guard here was mentioned in #37... but I'm pretty sure it was also the cause of a much earlier issue, #24, which unfortunately seems to have fallen through the cracks until now.

As for what a null means in this case: I believe the string in question ultimately comes from `System.IO.StreamReader.ReadLine()`, which will return null if the stream has been closed. I think the reason we don't see this every single time a stream connection is lost is that sometimes we will get an I/O error first and abandon the stream before we get here.

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSource.cs
+++ b/src/LaunchDarkly.EventSource/EventSource.cs
@@ -254,6 +254,12 @@ namespace LaunchDarkly.EventSource
 
         private void ProcessResponseContent(string content)
         {
+            if (content == null)
+            {
+                // StreamReader may emit a null if the stream has been closed; there's nothing to
+                // be done at this level in that case
+                return;
+            }
             if (string.IsNullOrEmpty(content.Trim()))
             {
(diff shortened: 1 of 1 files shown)


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -276,18 +282,18 @@ private void ProcessResponseContent(string content)
...
-                    return;
-                }
-                _readyState = state;
+            lock (this)
+            {
+                if (_readyState == state || _readyState == ReadyState.Shutdown)
+                {
+                    return;
+                }
+                _readyState = state;
```
eli-darkly: Spurious whitespace changes

eli-darkly:
I'd love to be able to unit-test this, but I'm not sure how because 1. I don't think we can control whether we will get the null and 2. I don't think the exception would be surfaced outside of the EventSource.

PR #22, opened 2018-02-20
Migrate to Common.Logging

atrakh:

PR #42, opened 2018-10-09
Replace Common.Logging with LibLog

derwasp:
_Warning: this change is not backwards compatible_

**Why did I create this PR?**

At [albelli](https://github.com/albumprinter/) we were using LaunchDarkly.Client of version [3.3.2](https://www.nuget.org/packages/LaunchDarkly.Client/3.3.2), which doesn't depend on [Common.Logging](https://github.com/net-commons/common-logging). Our application however does depend on it, but on a very old version, namely [2.2.0](https://www.nuget.org/packages/Common.Logging/2.2.0).
We are also using [log4net](https://logging.apache.org/log4net/) and a Common.Logging [connector](https://www.nuget.org/packages/Common.Logging.Log4Net1211/2.2.0) (note the old version number), which was not as restrictive as the [new version](https://github.com/net-commons/common-logging/blob/master/src/Common.Logging.Log4Net1211/Common.Logging.Log4Net1211.nuspec#L17). Due to legacy reasons, we have a bunch of different packages with slightly different versions of log4net. Assembly redirects solved the problem most of the time. But since your package depends on the new version of Common.Logging (and it should, since .NET core), the aforementioned connector should also be updated. But it requires only a specific log4net version.
All that results in us being unable to run our application if we update the SDK version.

**What does this PR do exactly?**

Since Common.Logging is a quite an old framework, which is not very actively maintained nowadays, I suggest to use [LibLog](https://github.com/damianh/LibLog) instead. It doesn't generate any additional dependencies, since it's installed as a source file.
The new version also has fancy installation way, but it doesn't support netstandard < 2.0, so I am installing it "the old way": by just copying the file in the source directory.

If you would consider dropping support for netstandard1.x, this file can be skipped and the PR would be more concise.

I already emailed your support team about that. But since we don't have a lot of time left, I decided that I might as well do a PR upfront.


derwasp:
One issue I see so far: 3 projects depend on common logging. And the dependency between the projects looks like this: EventSource <- Common <- Client
In order to use LibLog, we can:
a) Add it to bottom most library (EventSource in this case) and utilize [InternalsVisibleToAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netframework-4.7.2) to give other libraries access to `LogProvider.For<T>()` by using `LIBLOG_PUBLIC`. See [docu](https://github.com/damianh/LibLog/wiki/Using) for details.
b) We can add LibLog per library. (Which is kind of what this PR starts, since `LIBLOG_PUBLIC` is not defined here.
c) We can create a LaunchDarkly.Logging library which would only contain a `LibLog.cs` and `InternalsVisibleToAttribute` for the other libraries.

In any case, waiting for your feedback on this.

derwasp:
btw, I see that you actually used Microsoft.Extensions.Logging before.
What was the reason behind this PR?
https://github.com/launchdarkly/dotnet-eventsource/pull/22

eli-darkly:
> btw, I see that you actually used Microsoft.Extensions.Logging before.
> What was the reason behind this PR?

The reason for the change was briefly described in [issue #78](https://github.com/launchdarkly/dotnet-client/issues/78).

eli-darkly:
> One issue I see so far: 3 projects depend on common logging.

It's actually 4 - there's also our beta [Xamarin client](https://github.com/launchdarkly/xamarin-client/). This is a big concern I would have about putting in a new logging framework that's not actually shared between assemblies: if there is any significant change in it, then we would have to re-release all of those things, rather than just upgrading one dependency at the application level.

The other concern is of course just that this could require all our customers to once again change how their projects are set up. Although I'm not entirely sure what would happen if you took a project that currently had, say, Log4Net and Common.Logging, and then changed the library to use LibLog - would it just continue to work because it would auto-detect Log4Net?

In any case, I think this is something that will require more thought and we will probably also want to survey other developers in our community. The goal is to provide something that will work for, ideally, everyone.

eli-darkly:
I'm not sure what a `LaunchDarkly.Logging` library would accomplish since then we would either have to write connectors for it ourselves, or make our users write connectors, to support the existing logging frameworks that people are actually using.

As for "there should be no changes to this library" (LibLog), I am skeptical. If there is any change in the API of any of the logging frameworks that it supports, I don't see how it would be possible to avoid having to update it. It seems to me like very much the same issue as keeping the Common.Logging connectors up to date, except that it would be baked into our library code.

derwasp:
Let me start by [quoting Damian Hickey](https://github.com/damianh/LibLog/wiki), the author of LibLog here:

> LibLog is designed specifically and primarily for library developers that want to provide logging capabilities.

> This has been typically achieved in one of three ways:

> 1. **Take a hard dependency on a logging framework**. Thus forcing all consumers of your library to also take a dependency on your choice of logging framework. This causes issues when they want to use a different framework or are using a different version of the same one.
> 2. **Use Common.Logging**. Providing an abstraction over all the other frameworks has merit but it will add a reference to your consumers' projects. It's still a versionable dependency and can be prone to versioning clashes with other libraries. `Common.Logging's` `ILog` [contains ~65 members](https://github.com/net-commons/common-logging/blob/master/src/Common.Logging.Core/Logging/ILog.cs) whereas `LibLog's` `ILog` has just single method making it easier to implement custom loggers.
> 3. **Define your own Logging abstraction**. A common approach used by many libraries that don't want to do 1 or 2. `LibLog` is designed to help those that would otherwise go down this path.

So, the issue we have is the exact one that is described by Damian in `2.`. Since the library is not supposed to _write_ logs, but just have an option to connect any kind of logger, an interface is a way to solve this. LibLog provides exactly this interface.
It also has a buch of connectors to the different logging frameworks. But these should be considered supplementary, a nice addition to the basic interface. Because for sure not all the frameworks are available there. But when a new framework should be connected to _our library_ (LaunchDarkly Client in this case), the end consumer can just implement a very simple interface and everything will work. See [DebugLogProvider](https://github.com/launchdarkly/dotnet-eventsource/blob/3ef7b35f2e34f3ed2722b2c3c89132f378566b3e/console-app-dotnetcore/Program.cs#L12-L35) in my PR.

Also, in case an API of a specific logging library changes, I will quote Damian:

> Even though the APIs of the various loggers have been very stable over time, if they do change your library will fail gracefully. If (when!) that happens, users of your library can call LogProvider.SetCurrentLogProvider() to get things working again.

eli-darkly:
Am I correct that in your implementation, if you are using the LaunchDarkly SDK and want to specify a custom logging provider, you would have to specify it individually for each of the (3) assemblies in the SDK? If so, that is a problem as we really don't want people to have to worry about the exact internal structure of the project.

derwasp:
In this exact PR, yes. That is why I mentioned the option `c)`, in which a custom base logger package would be created and used for all the LaunchDarkly libraries. And customers would be able to hook into it.
To give other libraries access to it you have 2 options:

1. Use `InternalsVisibleToAttribute`
2. Set `LIBLOG_PUBLIC`

Please see my first comment for more details and links

PR #43, opened 2018-10-11
only reset backoff after a configurable threshold

eli-darkly:
This is one possible way to address the problem mentioned in issue #37, where if an event stream fails immediately after a successful connection, and this happens repeatedly, there is no increasing backoff delay. This is because we currently reset the backoff counter as soon as we get an HTTP 200.

The solution in this PR is to have a configurable time value, `BackoffResetThreshold`, defaulting to one minute. If the length of time between establishing a connection and the connection failing is less than this threshold, the backoff delay will continue to increase. If the connection lasted longer than the threshold, then the backoff delay is reset.

This is a fairly simple change, but I've also made a lot of changes in the unit tests (besides adding the tests for the new feature). The previous implementation of the mock server made it overly difficult to test reconnection scenarios, and also was not very realistic as it used complete responses rather than streaming responses. I hope this version makes it somewhat clearer what is going on.


.circleci/config.yml
```diff
@@ -16,4 +16,4 @@ jobs:
       - run: aws s3 cp s3://launchdarkly-pastebin/ci/dotnet/LaunchDarkly.EventSource.snk LaunchDarkly.EventSource.snk
       - run: dotnet restore
       - run: dotnet build src/LaunchDarkly.EventSource -f netstandard1.4
-      - run: dotnet test test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj -f netcoreapp1.1
+      - run: dotnet test test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj -f netcoreapp2.0
```
eli-darkly: I had to change the _test_ project to .NET Core 2.0 because the mock server code uses some I/O classes that aren't available in 1.1. This doesn't affect how the library itself is built.

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
```diff
@@ -22,12 +22,18 @@ public ExponentialBackoffWithDecorrelation(TimeSpan minimumDelay, TimeSpan maxim
         /// </summary>
         public TimeSpan GetNextBackOff()
         {
-            int nextDelay = Convert.ToInt32(Math.Min(_maximumDelay.TotalMilliseconds, _minimumDelay.TotalMilliseconds * Math.Pow(2, _reconnectAttempts)));
```
eli-darkly: I broke out this calculation into an `internal` method so that the tests can use it.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -16,64 +16,15 @@ public class EventSourceTests
         private readonly Uri _uri = new Uri("http://test.com");
         private readonly TimeSpan _defaultReadTimeout = TimeSpan.FromSeconds(1);
 
-        [Fact]
-        public void Exponential_backoff_should_not_exceed_maximum()
```
eli-darkly: These three tests are now in a separate file, because they don't test `EventSource`, they just test the backoff helper class.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -14,187 +14,95 @@ namespace LaunchDarkly.EventSource.Tests
...
-            evt.MessageReceived += (_, e) =>
-            {
-                message = e.Message;
-                wasMessageReceivedEventRaised = true;
-
-                evt.Close();
-            };
-
-            //// Act
+            handler.QueueResponse(StubResponse.Ok().WithAction(StubStreamAction.Write(sse)));
```
eli-darkly: For an explanation of why I added this "StubResponse/StubStreamAction" stuff instead of the simpler "QueueStringResponse", see StubHandler.cs.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -205,11 +113,13 @@ public async Task Default_HTTP_method_is_get()
         public async Task HTTP_method_can_be_specified()
         {
             var handler = new StubMessageHandler();
-            handler.QueueStringResponse(":");
+            handler.QueueResponse(StubResponse.Ok());
```
eli-darkly: Some tests like this were sending a dummy comment like ":" just in order to provide a trigger for closing the EventSource immediately. Instead I'm using the new `RequestReceived` event which fires as soon as the mock serve has received a request.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -14,187 +13,95 @@ namespace LaunchDarkly.EventSource.Tests
...
 
         [Fact]
         public async Task When_a_comment_SSE_is_received_then_a_comment_event_is_raised()
         {
-            // Arrange
-            var commentSent = ":";
+            var commentSent = ": hello";
 
             var handler = new StubMessageHandler();
+            handler.QueueResponse(StubResponse.StartStream(StreamAction.Write(commentSent + "\n\n")));
```
eli-darkly: For an explanation of why I added this "StubResponse/StreamAction" stuff instead of the simpler "QueueStringResponse", see StubHandler.cs.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -1,23 +1,25 @@
...
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace LaunchDarkly.EventSource.Tests
 {
     public class StubMessageHandler : HttpMessageHandler
```
eli-darkly: This is what's behind most of the test changes. The way this worked previously was that you would tell it how to respond to each HTTP request by providing a complete HTTP response. There were a couple problems with that: 1. each connection would be closed as soon as the response was sent, which is not an accurate simulation of how event streams normally behave, and also introduced reconnection behavior into every test whether we were testing that or not; 2. it provided no way to introduce a delay into a stream - the best you could do was use the "StubEventSource" class (now deleted) which hacked its own stream reading logic to pretend that there was a delay.

So, now, for successful responses that are supposed to provide a stream, it really provides streaming content and you can specify any number of writes and/or delays that should take place.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -26,41 +28,154 @@ public class StubMessageHandler : HttpMessageHandler
             if (_responses.Count == 0)
                 throw new InvalidOperationException("No response configured");
```
eli-darkly: This is a minor issue but I noticed that this logic doesn't exactly work - if it gets a request and there's no queued response, it will throw an exception, but that exception is just treated as a generic HTTP error. It would be nice if it could actually terminate the current unit test somehow, but I can't think how.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -26,41 +28,154 @@ public class StubMessageHandler : HttpMessageHandler
...
 
-            _requests.Add(request);
+            RequestReceived?.Invoke(this, request);
 
+            _requests.Add(request);
 
             var response = _responses.Dequeue();
-
-            /* Custom Response Message class provides the ability to throw an exception
-              This is useful for testing the retry capabilities.
```
eli-darkly: This is now done by specifying `StubResponse.WithIOError`.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -26,41 +28,154 @@ public class StubMessageHandler : HttpMessageHandler
...
+        public StubResponseWithStream(StreamAction[] actions)
+        {
+            _actions = actions;
+        }
+
+        override public HttpResponseMessage MakeResponse(CancellationToken cancellationToken)
+        {
+            var httpResponse = new HttpResponseMessage(HttpStatusCode.OK);
+            var streamRead = new AnonymousPipeServerStream(PipeDirection.In);
+            var streamWrite = new AnonymousPipeClientStream(PipeDirection.Out, streamRead.ClientSafePipeHandle);
```
eli-darkly: This is the part that required me to go up to .NET Core 2.0. It's possible that there is some other way to make a blocking stream that reads from another stream.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -26,41 +28,154 @@ public class StubMessageHandler : HttpMessageHandler
...
+                    {
+                        return;
+                    }
+                    byte[] data = Encoding.UTF8.GetBytes(action.Content);
+                    await output.WriteAsync(data, 0, data.Length, cancellationToken);
+                }
+                // if we've run out of actions, leave the stream open until it's cancelled
+                while (true)
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken);
```
eli-darkly: Something here that I'm unsure about: my expectation was that if the client cancels the request, `cancellationToken` will be cancelled, causing this line to throw an exception so that we'll exit the loop. However, in my debugging I was never able to see it get to line 143. I'm not sure if that means that I'm wrong about how the cancellation token works for HTTP requests, or if it just doesn't work because this is not a real HTTP server, or if it is somehow shutting down this task as soon as the unit test finishes so we never get that far.

So it is possible that these tasks are not stopping. I would rather not accumulate a bunch of old background tasks during a test run, just on general principle. On the other hand, they do go away as soon as the test run is done, so I'm not sure it is a huge problem.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
atrakh: If this is just an issue with the stub and is entirely encapsulated in the tests, could we just put an assert inside of the `SendAsync` stub?

eli-darkly: To be clear, the issue is that if the stub server receives 2 HTTP requests, but we had only set it up with 1 response, then something is wrong with the test and it should probably die. But I don't think using `Assert` here would do anything more than `throw` is doing. An assertion failure just causes an exception to be thrown, that's how it's implemented.

PR #45, opened 2018-12-13
Remove unnecessary dependencies

j3parker:


console-app/EventSource-ConsoleApp.csproj
```diff
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0"
-  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
j3parker: Visual Studio made this change

console-app/EventSource-ConsoleApp.csproj
```diff
@@ -38,12 +37,6 @@
     <Reference Include="System.Core" />
     <Reference Include="System.IO.Compression" />
     <Reference Include="System.Numerics" />
-    <Reference Include="System.Runtime.CompilerServices.Unsafe, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
j3parker: This one was unused and VS was complaining about the version number mismatch here.
I noticed it when doing an audit of DLLs that we ship at my company.

console-app/EventSource-ConsoleApp.csproj
```diff
@@ -38,12 +37,6 @@
     <Reference Include="System.Core" />
     <Reference Include="System.IO.Compression" />
     <Reference Include="System.Numerics" />
-    <Reference Include="System.Runtime.CompilerServices.Unsafe, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\packages\System.Runtime.CompilerServices.Unsafe.4.3.0\lib\netstandard1.0\System.Runtime.CompilerServices.Unsafe.dll</HintPath>
-    </Reference>
-    <Reference Include="System.Runtime.InteropServices.RuntimeInformation, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
j3parker: This one is transitively referenced by `Microsoft.CSharp`. VS was complaining about the version number mismatch.

j3parker:
@eli-darkly bump?

eli-darkly:
@j3parker Sorry! I hadn't seen your PR - it looks like I accidentally turned off notifications somehow. So thanks for reminding me. This looks right, I'll just make sure it doesn't affect anything in our various test environments.

console-app/EventSource-ConsoleApp.csproj
eli-darkly: I'm a bit confused about that last part since this project isn't part of anything that would be shipped; it's strictly for manual testing of the library. The dependency in question does seem to be unnecessary, but I'm not sure how it would have any effect on what you're doing.

eli-darkly: Though I am leaning toward just getting rid of this test app—I think it was only in there because at first we didn't have any real end-to-end HTTP in the unit tests.

j3parker:
Cool! No problem. Thanks.

eli-darkly:
After some thought, it seems like that test app isn't serving any purpose. It'll go away in the next release, which should be pretty soon. Thanks for pointing it out.

eli-darkly:
They've been removed in the latest release.

PR #47, opened 2019-01-08
add dotnet-base release script and digital signing; don't strong-name in Debug mode

eli-darkly:
This updates the repo to our current build standard. Release artifacts will now actually be built in the Release configuration, and will also be signed as well as strong-named. The default Debug configuration does neither of those things.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the LaunchDarkly's EventSource implementation for C# will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
 
+## [3.2.1] - 2019-01-07
```
atrakh: does the date need to be updated?

eli-darkly: Yes, but I'll update it when I actually do a release.

PR #50, opened 2019-08-07
clean up connection/client/error management, add debug logging

eli-darkly:
This is a step toward a larger reorganization of the EventSource code that I'm hoping will improve both readability and reliability. I noticed while trying to diagnose a connection problem that it was a bit hard to follow how errors and connection state are being managed. See comments for details.

I also added more debug-level logging; I might remove some of this later but right now it's helpful in verifying that the logic is working as intended.

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSource.cs
+++ b/src/LaunchDarkly.EventSource/EventSource.cs
@@ -1,6 +1,7 @@
 ﻿using Common.Logging;
 using System;
 using System.Collections.Generic;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -12,12 +13,12 @@ namespace LaunchDarkly.EventSource
     /// </summary>
     public class EventSource : IEventSource, IDisposable
     {
-
         #region Private Fields
 
         private readonly Configuration _configuration;
         private readonly ILog _logger;
 
+        private HttpClient _httpClient;
         private List<string> _eventBuffer;
         private string _eventName = Constants.MessageField;
         private string _lastEventId;
@@ -123,6 +124,11 @@ namespace LaunchDarkly.EventSource
         /// <exception cref="InvalidOperationException">The method was called after the connection <see cref="ReadyState"/> was Open or Connecting.</exception>
         public async Task StartAsync()
         {
+            if (_httpClient is null)
+            {
+                _httpClient = CreateHttpClient();
+            }
+
             bool firstTime = true;
             while (ReadyState != ReadyState.Shutdown)
             {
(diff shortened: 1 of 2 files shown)


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -12,12 +13,12 @@ namespace LaunchDarkly.EventSource
     /// </summary>
     public class EventSource : IEventSource, IDisposable
     {
-
         #region Private Fields
 
         private readonly Configuration _configuration;
         private readonly ILog _logger;
 
+        private HttpClient _httpClient;
```
eli-darkly: This is the biggest functional change, especially in the context of the Xamarin SDK. Previously, we were creating a new `HttpClient` for every connection attempt. In server-side .NET that may not be a big deal, but in a mobile environment we do a lot more reconnecting, and since the client properties are always the same I decided to just use one of them and move it up here.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -138,28 +144,62 @@ public async Task StartAsync()
...
+                }
+
+                if (ReadyState == ReadyState.Connecting || ReadyState == ReadyState.Open)
+                {
+                    throw new InvalidOperationException(string.Format(Resources.EventSource_Already_Started, ReadyState));
+                }
+
+                SetReadyState(ReadyState.Connecting);
+                cancellationToken = newRequestTokenSource.Token;
+
```
eli-darkly: This logic is more or less equivalent to what we had, but some of it was down in the `ConnectToEventSourceAsync` method, and I pulled it up here because it wasn't that closely related to the main task of that method (setting up the `EventSourceService`). We may not even need that method— it's never used anywhere else.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -138,28 +144,70 @@ public async Task StartAsync()
...
+                        }
+
+                        if (realException != null)
+                        {
+                            _logger.ErrorFormat("Encountered an error connecting to EventSource: {0}", realException, realException.Message);
+                            _logger.Debug("", realException);
+
+                            OnError(new ExceptionEventArgs(realException));
+                        }
+                    }
```
eli-darkly: This b lock is just a consolidation of the error-handling logic with code that used to be in `ConnectToEventSourceAsync`. We were handling some errors down there and rethrowing others up here, for no particular reason. I also added the logic on lines 188-200 that detects timeouts (which in .NET are not actually thrown as `TimeoutException` but as `OperationCanceledException` for some reason) and avoids showing stacktraces for pointless exceptions that we deliberately caused.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -180,10 +228,13 @@ public async Task StartAsync()
         /// </summary>
         public void Close()
         {
-            if (ReadyState == ReadyState.Raw || ReadyState == ReadyState.Shutdown) return;
```
eli-darkly: Removed this `return` because we do want to make sure we dispose of the HTTP client even if we're not currently active.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -138,28 +144,67 @@ public async Task StartAsync()
...
+                    await ConnectToEventSourceAsync(cancellationToken);
+
+                    // ConnectToEventSourceAsync normally doesn't return, unless it detects that the request has been cancelled.
+                    Close(ReadyState.Closed);
                 }
                 catch (Exception e)
-                {
-                    _logger.ErrorFormat("Encountered an error connecting to EventSource: {0}", e, e.Message);
-                    _logger.Debug("", e);
-                }
```
eli-darkly: I removed this ERROR-level logging because we are already giving the caller a way to be notified of errors, so they can easily decide whether they want to make those errors so prominent in their logs or not. ERROR level can have specific consequences for system monitoring, so in a very general-purpose library like this I think it's best not to make assumptions about how they want such things reported. In the context of the LaunchDarkly SDK, we are already doing our own error logging there in response to the Error event, so we were getting everything twice.

PR #51, opened 2019-08-08
Don't create a default HttpMessageHandler, + misc config cleanup

eli-darkly:
We've been always using the `HttpClient` constructor that can take an `HttpMessageHandler` instance, and if the application didn't specify one then we create one. There's really no reason to do this unless we actually want to override the message handler, and [on some platforms](https://nicksnettravels.builttoroam.com/post-2019-04-24-xamarin-and-the-httpclient-for-ios-android-and-windows-aspx/) it prevents the OS from substituting a better implementation.

I also did some general cleanup of the configuration classes, mostly just doc comments.

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSource.cs
+++ b/src/LaunchDarkly.EventSource/EventSource.cs
@@ -16,9 +16,9 @@ namespace LaunchDarkly.EventSource
         #region Private Fields
 
         private readonly Configuration _configuration;
+        private readonly HttpClient _httpClient;
         private readonly ILog _logger;
 
-        private HttpClient _httpClient;
         private List<string> _eventBuffer;
         private string _eventName = Constants.MessageField;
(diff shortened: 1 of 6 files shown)


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -124,11 +125,6 @@ public EventSource(Configuration configuration)
         /// <exception cref="InvalidOperationException">The method was called after the connection <see cref="ReadyState"/> was Open or Connecting.</exception>
         public async Task StartAsync()
         {
-            if (_httpClient is null)
-            {
-                _httpClient = CreateHttpClient();
-            }
-
```
eli-darkly: I'm not sure why I was creating this lazily before. It's simpler to do it in the constructor and then if our parameters are invalid, we'll find out right away.

PR #53, opened 2019-10-17
Set LastEventId header on reconnection

fracek:
This patch fixes a problem that arises when disconnecting and then reconnecting.
If the `last-event-id` header is not set, the server can send old events. By setting the header to the the last event id seen, we ensure we always receive up-to-date events.

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSourceService.cs
+++ b/src/LaunchDarkly.EventSource/EventSourceService.cs
@@ -63,24 +63,24 @@ namespace LaunchDarkly.EventSource
         /// Initiates the request to the EventSource API and parses Server Sent Events received by the API.
         /// </summary>
         /// <returns>A <see cref="System.Threading.Tasks.Task"/> A task that represents the work queued to execute in the ThreadPool.</returns>
-        public async Task GetDataAsync(Action<string> processResponse, CancellationToken cancellationToken)
+        public async Task GetDataAsync(Action<string> processResponse, string lastEventId, CancellationToken cancellationToken)
         {
             cancellationToken.ThrowIfCancellationRequested();
 
-            await ConnectToEventSourceApi(processResponse, cancellationToken);
+            await ConnectToEventSourceApi(processResponse, lastEventId, cancellationToken);
         }
 
         #endregion
 
         #region Private Methods
 
-        private async Task ConnectToEventSourceApi(Action<string> processResponse, CancellationToken cancellationToken)
+        private async Task ConnectToEventSourceApi(Action<string> processResponse, string lastEventId, CancellationToken cancellationToken)
         {
             _logger.DebugFormat("Making {0} request to EventSource URI {1}",
                 _configuration.Method ?? HttpMethod.Get,
                 _configuration.Uri);
-            
-            using (var response = await _httpClient.SendAsync(CreateHttpRequestMessage(_configuration.Uri),
+
+            using (var response = await _httpClient.SendAsync(CreateHttpRequestMessage(_configuration.Uri, lastEventId),
                 HttpCompletionOption.ResponseHeadersRead,
                 cancellationToken).ConfigureAwait(false))
             {
@@ -140,8 +140,8 @@ namespace LaunchDarkly.EventSource
                 processResponse(line);
             }
         }
-        
-        private HttpRequestMessage CreateHttpRequestMessage(Uri uri)
+
+        private HttpRequestMessage CreateHttpRequestMessage(Uri uri, string lastEventId)
         {
             var request = new HttpRequestMessage(_configuration.Method ?? HttpMethod.Get, uri);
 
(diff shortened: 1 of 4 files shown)


src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -164,6 +164,10 @@ private HttpRequestMessage CreateHttpRequestMessage(Uri uri)
                 }
             }
 
+            // If the lastEventId was provided, include it as a header to the API request.
+            if (!string.IsNullOrWhiteSpace(lastEventId) && !request.Headers.Contains(Constants.LastEventIdHttpHeader))
+                request.Headers.Add(Constants.LastEventIdHttpHeader, lastEventId);
+
```
eli-darkly: I think this is _slightly_ wrong, in that it will _not_ override any `Last-Event-Id` header that was provided in the configuration.

Here's a scenario:

* The application creates an EventSource, and specifies in the configuration that it should start with a header of `Last-Event-Id: 123`. So, on the first connection attempt, it sends that header.
* It then receives several events from the stream, and they each have an `id` field. The last one received has `id: 131`. So the private field `_lastEventId` is now "131".
* Now there's a connection error, so we need to reconnect.

I think that in this scenario, it would be correct for it to send `Last-Event-Id: 131`, not `Last-Event-Id: 123`. The configuration specified what the application thought the last event was at _that_ time, but the state has changed since then. So I think on lines 168-169, it would be appropriate for it to overwrite any previously specified value for this header.

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -164,6 +164,10 @@ private HttpRequestMessage CreateHttpRequestMessage(Uri uri)
                 }
             }
 
+            // If the lastEventId was provided, include it as a header to the API request.
+            if (!string.IsNullOrWhiteSpace(lastEventId) && !request.Headers.Contains(Constants.LastEventIdHttpHeader))
+                request.Headers.Add(Constants.LastEventIdHttpHeader, lastEventId);
```
eli-darkly: I think this is _slightly_ wrong, in that it will _not_ override any `Last-Event-Id` header that was provided in the configuration.

Here's a scenario:

* The application creates an EventSource, and specifies in the configuration that it should start with a header of `Last-Event-Id: 123`. So, on the first connection attempt, it sends that header.
* It then receives several events from the stream, and they each have an `id` field. The last one received has `id: 131`. So the private field `_lastEventId` is now "131".
* Now there's a connection error, so we need to reconnect.

I think that in this scenario, it would be correct for it to send `Last-Event-Id: 131`, not `Last-Event-Id: 123`. The configuration specified what the application thought the last event was at _that_ time, but the state has changed since then. So I think on lines 168-169, it would be appropriate for it to overwrite any previously specified value for this header.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -192,6 +192,39 @@ public async Task When_LastEventId_is_configured_then_the_outgoing_request_conta
...
+            };
+
+            await evt.StartAsync();
+
+            var requests = handler.GetRequests().ToList();
+            Assert.False(requests[0].Headers.Contains(Constants.LastEventIdHttpHeader));
+            Assert.True(requests[1].Headers.Contains(Constants.LastEventIdHttpHeader));
+            Assert.True(requests[1].Headers.GetValues(Constants.LastEventIdHttpHeader).Contains(lastEventId));
+        }
+
```
eli-darkly: As per the previous comment, I would add a second version of this test in which the original configuration specifies a different last event ID.

src/LaunchDarkly.EventSource/EventSourceService.cs
fracek: Good catch.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
fracek: I added a test where it sets both the `LastEventId` configuration option and manually adds a `Last-Event-Id` header. The behaviour you mentioned was only triggered when adding the header.

fracek:
Thank you for taking the time to review this. I updated the PR.

src/LaunchDarkly.EventSource/EventSourceService.cs
eli-darkly: I don't think that calling `request.Headers.Add` will _overwrite_ an existing value, though. It will add a second value to the header, which we don't want.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.namevaluecollection.add?view=netframework-4.8

fracek: You are correct. I also added a check in the test to catch this behaviour.

eli-darkly:
This is in the 3.3.2 release.

PR #55, opened 2019-11-15
await in Can_Create_and_start_EventSource_without_specifying_message_…

vslee:
…handler() test

### DIFF
--- a/test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
+++ b/test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
@@ -15,11 +15,11 @@ namespace LaunchDarkly.EventSource.Tests
         private readonly Uri _uri = new Uri("http://test.com");
 
         [Fact]
-        public void Can_Create_and_start_EventSource_without_specifying_message_handler()
+        public async Task Can_Create_and_start_EventSource_without_specifying_message_handler()
         {
             // Testing this just because all of the other tests use a StubMessageHandler
             var evt = new EventSource(Configuration.Builder(_uri).Build());
-            evt.StartAsync();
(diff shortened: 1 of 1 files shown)


test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -15,11 +15,11 @@ public class EventSourceTests
         private readonly Uri _uri = new Uri("http://test.com");
 
         [Fact]
-        public void Can_Create_and_start_EventSource_without_specifying_message_handler()
+        public async Task Can_Create_and_start_EventSource_without_specifying_message_handler()
         {
             // Testing this just because all of the other tests use a StubMessageHandler
             var evt = new EventSource(Configuration.Builder(_uri).Build());
-            evt.StartAsync();
+			await evt.StartAsync();
```
eli-darkly: I don't think this will work, because the task returned `StartAsync` does not complete unless you shut down the EventSource. Have you run the tests with this change? I would expect the test to hang.

(I see that we hadn't enabled automatic CI builds for pull requests in this repo, so I've enabled them now.)

vslee: You're right, it does hang. I'll close this.

eli-darkly: Well, it's also not good that we are starting an EventSource there and never stopping it.

I'm not sure this test needs to exist at all; the fact that you can call the constructor and not get an exception is already covered by basically all of the other tests. The test code is pretty old and messy and I've been meaning to take a pass at cleaning it up, so that might include getting rid of this. Thanks for bringing it up.

PR #56, opened 2019-11-15
enable users to run tests locally

vslee:
- published version is signed
- however, users don't have the key to sign the project locally
- the assembly info prev only allowed testing from a signed EventSource.Tests
- now, allows building tests in debug mode w/o signing, and allows access from unsigned EventSource.Tests
- theoretically possible for outsiders to create an unrelated EventSource.Tests to get internal access to EventSource, but those variables are already accessible through reflection

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,6 @@ jobs:
           name: install packages
           command: apt-get -q update && apt-get install -qy awscli
       - checkout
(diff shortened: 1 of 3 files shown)


eli-darkly:
Could you try removing the `aws s3 cp` command from `.circleci/config.yml` - which can't work in a fork build, but was only there because the tests required a key - and see if that makes the CI build pass?

PR #61, opened 2020-11-13
Usage of custom httpClient

thorstenfleischmann:
This enables the usage of an existing httpClient. This can be useful if you can not provide an apropriate MessageHandler. In my specific case I have to reuse an exiting httpClient with some included security handlers. This existing httpClient is provided by a (external) factory so i can not provide the handlers to the eventsource myself.

### DIFF
--- a/src/LaunchDarkly.EventSource/Resources.resx
+++ b/src/LaunchDarkly.EventSource/Resources.resx
@@ -123,6 +123,12 @@
   <data name="Configuration_Value_Greater_Than_Zero" xml:space="preserve">
     <value>Value must be greater than zero.</value>
   </data>
+  <data name="Configuration_HttpClient_With_MessageHandler" xml:space="preserve">
+    <value>HttpClient can not be used with MessageHandler.</value>
+  </data>
+  <data name="Configuration_HttpClient_With_ConnectionTimeout" xml:space="preserve">
+    <value>HttpClient can not be used with ConnectionTimeout. Use Timeout property of httpClient instead.</value>
+  </data>
   <data name="EventSourceService_Read_Timeout" xml:space="preserve">
     <value>A timeout occurred while reading the request from the remote EventSource API</value>
   </data>
--- a/src/LaunchDarkly.EventSource/Configuration.cs
+++ b/src/LaunchDarkly.EventSource/Configuration.cs
@@ -84,6 +84,11 @@ namespace LaunchDarkly.EventSource
         public HttpMessageHandler MessageHandler { get; }
 
         /// <summary>
+        /// The HttpClient that will be used as the HTTP client, or null for a new HttpClient.
+        /// </summary>
+        public HttpClient HttpClient { get; }
+
+        /// <summary>
         /// The HTTP method that will be used when connecting to the EventSource API.
         /// </summary>
         public HttpMethod Method { get; }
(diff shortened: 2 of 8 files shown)


src/LaunchDarkly.EventSource/Resources.resx
```diff
@@ -1,153 +1,159 @@
...
-  <data name="EventSource_Logger_Connection_Error" xml:space="preserve">
-    <value>Encountered exception in EventSourceService.ConnectToEventSourceApi method. Exception Message: {0}</value>
-  </data>
-  <data name="EventSource_Logger_Disconnected" xml:space="preserve">
-    <value>EventSource Disconnected. Automatically delaying {0}ms before reconnecting.</value>
-  </data>
-  <data name="EventSource_Response_Content_Empty" xml:space="preserve">
-    <value>Http Response contained no content.</value>
-  </data>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
eli-darkly, line 1: Please don't check in changes that touch the entire file when it's only because your IDE decided to reformat them. That makes it impossible to track what was really changed in the git history.

eli-darkly:
Thanks for the contribution, especially for making the effort to add tests. This looks good at first glance but I'll need to return to look at it in more detail a little later. Do you have an immediate time frame for needing this, or are you able to use a fork for now? I ask because we're in the middle of extensive rewrites on our .NET libraries, including this one, and it might be easier for us to make a change similar to this in the new version, rather than merge this PR and then bring the changes forward through a bunch of merge conflicts.

thorstenfleischmann:
It would be nice to have this feature within a few weeks. 

Apart from the configuration the changes boil down to two lines of new logic (construction of event source httpclient and exclude from dispose)

What's the goal of the rewrite? I think the code is very clean and easy to use.

eli-darkly:
> What's the goal of the rewrite? I think the code is very clean and easy to use.

There are several— some that are related to how we use this library within our .NET SDK, and some that aren't.

1. Most of the code is quite old, and it was originally required to be compatible with .NET Standard 1.x. We're raising the lowest compatible platform to .NET Standard 2.0 / .NET Core 2.1 / .NET Framework 4.5.2 and there may be some implementation changes related to that.
2. The error handling flow is hard to follow and we've seen some intermittent failure modes that are hard to diagnose. We believe this can be improved.
3. The unit test coverage is spotty and it's implemented in several inconsistent ways, due to some of it being inherited when we forked from an earlier codebase.
4. The dependency on Common.Logging is undesirable for some users. Because our .NET libraries can be deployed in a wide variety of platforms (Windows, .NET Core on Linux, Xamarin) there isn't really a single logging framework or facade that wouldn't be problematic on some of them. So we'll be replacing this with a simple logging abstraction that can be made to delegate to Common.Logging, Microsoft.Extensions.Logging, log4net, or other popular frameworks with a single line of configuration.

eli-darkly:
If you're curious, you can follow that work as it happens on the `4.x` branch. Once that's ready, it'll be merged to master and the current master will become a 3.x maintenance branch.

Anyway, I was not complaining about this being a major change, it's pretty simple, it's just that if you really didn't care when it happened then we might put it off until 4.0 just to avoid one more merge-forward and merge conflict since the 4.x branch is already under way. But that's not a big deal and I think we should be able to merge and release this within a week.

thorstenfleischmann:
Thanks for the detailled explanations and feedback. I really appreciate it! 

Another idea would be to simply make CreateHttpClient() and Close() "public override" so I can create my own event source class.

eli-darkly:
> Another idea would be to simply make CreateHttpClient() and Close() "public override" so I can create my own event source class. (Not completely sure if this would work)

It would sort of work, although overriding Close() would not be good because there's other behavior in Close() unrelated to the HTTP client. So if we did something like this, we would instead provide a more specific method like CloseHttpClient that you could override.

But, in general we prefer to implement things like this through configuration or composition, rather than inheritance. Letting a complex class be subclassed can get messy pretty fast.

eli-darkly:
I'm changing the target branch to `contrib` so we can make some slight changes before it goes into `master`, but this will be in the 3.4.0 release.

eli-darkly:
Version 3.4.0 is out now with this feature. Thanks again.

PR #63, opened 2021-01-16
add option to read UTF8 stream as raw bytes

eli-darkly:
NOTE: `master` is now for the upcoming major version 4.0. The previous `master` code is now on a `3.x` maintenance branch.

Previously, EventSource always used the .NET `StreamReader` to read incoming HTTP response data and convert it transparently to the .NET `string` type, using whatever character encoding was present in the response.

That has a disadvantage for applications that expect to receive potentially large amounts of data in each event and that would like to process it as a UTF-8-encoded byte array rather than a `string`. An example of this case is the upcoming LaunchDarkly .NET SDK 6.0, which (on most platforms) will use the `System.Text.Json` API to parse JSON data from stream events; that API expects the input to be UTF-8 bytes, so if what you have is a `string`, you have to copy the whole thing into a new byte buffer using `Encoding.GetBytes()`. And, during the time in between EventSource reading the data and the application processing it, it's using a lot more memory as necessary, because `string` uses UTF-16 characters with a minimum of two bytes per character. That excessive transient memory usage is an existing issue in current versions of the LaunchDarkly SDK: if you receive an event containing 100K characters of JSON data, assuming there are no multi-byte characters, it's allocating at least 200K to hold that data until it's been processed.

This PR adds an opt-in mode for greater efficiency in such a use case. If you set `ConfigurationBuilder.PreferDataAsUtf8Bytes(true)`, _and_ if the stream's encoding actually is UTF-8, then the `data` field of a stream event will be read into a byte buffer with no changes (except changing CR or CRLF line endings to LF, per the SSE spec), and that buffer will be passed to the message handler in the new property `MessageEvent.DataUtf8Bytes`. (If you read the old string property `MessageEvent.Data` instead, it will still work, but it will do a bytes->string conversion at that point.) EventSource does not use `StreamReader` in this case; instead it implements its own buffering and linefeed-scanning logic.

In all other cases— if the stream's encoding is not UTF-8, or if you didn't set `PreferDataAsUtf8Bytes`— EventSource behaves exactly as before: messages are parsed and stored as strings. (And if you read `MessageEvent.DataUtf8Bytes` in that case, it will still work, you'll just have the overhead of a string->bytes conversion.)

I also added a `DefaultEncoding` configuration property for the unlikely case that the server doesn't say what the encoding is and you want it to default to something that's _not_ UTF-8.

Unrelated: as part of the rewrite of the stream reading logic, I noticed that we were using an inefficient method of enforcing a read timeout, where we always created an extra task with `Task.Delay()` and checked whether it finished executing before our read task. In most .NET platforms there's now a lower-level way to set read timeouts at the socket level, so we can now try to use that and fall back to the old way if that API isn't available.


src/LaunchDarkly.EventSource/EventParser.cs
```diff
@@ -9,119 +10,87 @@ namespace LaunchDarkly.EventSource
...
-        /// <summary>
-        /// Determines if the specified value is a data field in a Server Sent Event message.
-        /// </summary>
-        /// <param name="value">The value.</param>
-        /// <returns>
-        ///   <c>true</c> if the specified value is a data field; otherwise, <c>false</c>.
-        /// </returns>
-        public static bool IsDataFieldName(string value)
-        {
-            return Constants.DataField.Equals(value, StringComparison.OrdinalIgnoreCase);
```
eli-darkly, line 36: The logic for all of these `Is...FieldName` methods was incorrect: they were doing a case-insensitive comparison, but the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) says "Field names must be compared literally, with no case folding performed." I've changed this so it just does regular string comparisons in the `Is...` properties of `Result`.

That's a breaking change, sort of, in that it means EventSource will fail to process data that it would've previously accepted. But since that is by definition not valid stream data at all, it's more of a bug fix.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -354,43 +361,85 @@ private void SetReadyState(ReadyState state, Action<StateChangedEventArgs> actio
...
             {
-                _lastEventId = value;
+                if (result.ValueString != null)
+                {
+                    if (_eventDataStringBuffer is null)
+                    {
+                        _eventDataStringBuffer = new List<string>(2);
+                    }
+                    _eventDataStringBuffer.Add(result.ValueString);
+                    _eventDataStringBuffer.Add("\n");
```
eli-darkly, line 379: It may seem a bit odd to use an array of strings rather than for instance a `StringBuilder`, but the previous author was actually being clever in this case (in the previous version of this same logic on lines 361-362). It's likely that most events have only a single line of `data:`. In that case, after allocating the initial `List<string>`, it'll simply take the `string` instance for the line that it's just parsed and put that in as element 0, and then put the literal `"\n"` in as element 1; then, when it realizes there are no more lines, it'll call `string.Concat` on just the first element, which just returns the existing `string`. And then the same `List` will be reused, so it's avoiding unnecessary allocations which would happen if we were appending to a buffer.

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -91,53 +106,130 @@ private async Task ConnectToEventSourceApi(Action<string> processResponse, strin
...
                 using (var stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false))
                 {
-                    using (var reader = GetStreamReader(stream))
+                    bool needManualTimeout = false;
+                    try
+                    {
+                        // Some Stream subclasses on some platforms have their own read timeout implementation; others
+                        // don't. The only way to find out is to try to set it and see if you get an exception. If we
+                        // get an exception, then ReadTimeout won't work and we'll need to implement our own timeout.
+                        stream.ReadTimeout = (int)_configuration.ReadTimeout.TotalMilliseconds;
```
eli-darkly, line 115: This is the improved read timeout logic I mentioned in the PR description.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -570,11 +583,12 @@ public async Task When_error_handler_closes_event_source_no_reconnect_attempt_is
             Assert.Equal(1, handler.GetRequests().Count());
         }
 
+        [Fact]
         public async Task Connection_closed_before_threshold_gets_increasing_backoff_delay()
```
eli-darkly, line 587: The fixes in this test and the next test are unrelated to this PR, but these tests were broken and always had been. We just didn't see it because they weren't actually being run (due to the lack of `[Fact]`).

test/LaunchDarkly.EventSource.Tests/Stubs/EventSink.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
+using System.Collections.Concurrent;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace LaunchDarkly.EventSource.Tests
+{
+    public class EventSink
```
eli-darkly, line 8: We can use this class and some of the other approaches I used in the newer tests to streamline the existing tests quite a bit, but I'll save that for another PR.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -568,98 +581,6 @@ public async Task When_error_handler_closes_event_source_no_reconnect_attempt_is
             await evt.StartAsync();
 
             Assert.Equal(1, handler.GetRequests().Count());
-        }
-
-        public async Task Connection_closed_before_threshold_gets_increasing_backoff_delay()
```
eli-darkly, line 573: Unrelated to this PR: I removed this test and the following test because they were fatally flawed and guaranteed to be unstable. The only reason we weren't seeing frequent intermittent failures on them in CI was that... they weren't actually being run at all, due to the lack of a `[Fact]` annotation.

The problem is that these tests were trying to measure the timing of actions being done in a live HTTP session between client and server. Using actual I/O makes the execution time completely unpredictable; for that matter, even without I/O, task scheduling in .NET is unpredictable enough that even a simple test involving multiple tasks will have wildly variable execution time.

So, while we of course can and do test the lower-level `ExponentialBackoffWithDecorrelation` logic separately, if we really want to verify that `EventSource` is _using_ the backoff logic as intended, we would need to add some more abstractions/instrumentation inside `EventSource` to make it possible to observe what it _intends_ the backoff delays to be, rather than what they really end up being.

src/LaunchDarkly.EventSource/ByteArrayLineScanner.cs
```diff
@@ -0,0 +1,157 @@
...
+            if (_partialLine != null && _partialLine.Position > 0)
+            {
+                _partialLine.Write(_buffer, startedAt, pos - startedAt);
+                lineOut = new Utf8ByteSpan(_partialLine.GetBuffer(), 0, (int)_partialLine.Position);
+                _partialLine = null;
+
+                // If there are still bytes in the buffer, move them over to make more room
+                if (_startPos < _count)
+                {
+                    System.Buffer.BlockCopy(_buffer, _startPos, _buffer, 0, _count - _startPos);
```
gwhelanLD, line 141: Is there a thought process here for doing this shift copy in the case where there was a `_partialLine` vs the reset when the buffer bytes ended on a line ending?

Thinking about it more, I would guess that's because we normally expect from the SSE stream. I guess it makes sense to me then, but took a fair bit of thought.

src/LaunchDarkly.EventSource/EventParser.cs
```diff
@@ -9,119 +10,87 @@ namespace LaunchDarkly.EventSource
...
+            internal bool IsDataField => FieldName == "data";
+            internal bool IsEventField => FieldName == "event";
+            internal bool IsIdField => FieldName == "id";
+            internal bool IsRetryField => FieldName == "retry";
         }
 
         /// <summary>
-        /// Gets the field and value from a server sent event message.
+        /// Attempts to parse a single non-empty line of SSE content that was read as a string. Empty lines
+        /// shoudl not be not passed to this method.
```
gwhelanLD, line 33: typo: `shoudl`

src/LaunchDarkly.EventSource/EventParser.cs
```diff
@@ -9,119 +10,87 @@ namespace LaunchDarkly.EventSource
...
+            {
+                FieldName = line.Substring(0, colonPos),
+                ValueString = line.Substring(valuePos)
+            };
         }
 
         /// <summary>
-        /// Determines if the specified string is numeric (contains only numeric characters).
+        /// Attempts to parse a single non-empty line of SSE content that was read as UTF-8 bytes. Empty lines
+        /// shoudl not be not passed to this method.
```
gwhelanLD, line 63: typo: `shoudl`

src/LaunchDarkly.EventSource/MessageEvent.cs
```diff
@@ -46,87 +106,110 @@ public MessageEvent(string data, Uri origin) : this(data, null, origin)
...
+
+        /// <summary>
+        /// Gets the data received in the event as a UTF-8 byte span.
+        /// </summary>
+        /// <remarks>
+        /// If the data was originally provided as UTF-8 bytes, the returned value refers to
+        /// the same array, offset, and length (it is the caller's responsibility not to
+        /// modify the byte array). If it was originally provided as a string, the string
+        /// is copied to a new byte array.
+        /// </remarks>
```
gwhelanLD, line 128: Maybe should have a `<value>` tag.

src/LaunchDarkly.EventSource/MessageEvent.cs
```diff
@@ -46,87 +106,110 @@ public MessageEvent(string data, Uri origin) : this(data, null, origin)
...
+        /// <summary>
+        /// True if the event data is stored internally as UTF-8 bytes.
+        /// </summary>
+        /// <remarks>
+        /// The data can be accessed with either <see cref="Data"/> or <see cref="DataUtf8Bytes"/>
+        /// regardless of the value of this property. The property only indicates the <i>original</i>
+        /// format of the data, so, for instance, if it is <see langword="true"/> then
+        /// reading <see cref="Data"/> will have more overhead (due to copying) than
+        /// reading <see cref="DataUtf8Bytes"/>.
+        /// </remarks>
```
gwhelanLD, line 157: Same as above for not having a `<value>` tag. That said, I've not looked at the generated code doc for this API, I'm not sure how important it is to have that tag.

gwhelanLD (changes requested):
Some nitpicks on comments, but otherwise looks good.

src/LaunchDarkly.EventSource/MessageEvent.cs
eli-darkly: I've never been convinced that there is any point to using the `<value>` tag. You end up with a bunch of doc comments that look like this:

```
/// <summary>The number of items.</summary>
/// <value>the number of items</value>
int Count { get; }
```

I mean, it's a property, it doesn't have any parameters, so I can't think of any circumstance in which the `<value>` text wouldn't just be exactly the same as the `<summary>`— or a slightly shorter version of, but why would that matter when the `<summary>` is what you see _first_ in the docs?

gwhelanLD: Yeah, I've been basically thinking the same, I just hadn't looked at the generated docs to see if it seemed important to how they turn out. Do you know if it provides any value in IDEs that `summary` does not cover?

eli-darkly: No, I've never found any reason at all that it's helpful. IDEs that show you a doc comment as hover text, etc., will show the `<summary>`.

src/LaunchDarkly.EventSource/ByteArrayLineScanner.cs
eli-darkly: I'm not sure I understand the question; can you be more specific about which code path you mean by "the reset when the buffer bytes ended on a line ending"? If you mean line 151 - the answer is in the question. That is, if we have no more bytes in the buffer, then there's nothing to shift over.

src/LaunchDarkly.EventSource/MessageEvent.cs
gwhelanLD: Alright then, I'll keep that in mind when reading dotnet doc comments then. They just stood out as an inconsistency to me.

eli-darkly: I mean, to me, even the official documentation for this tag doesn't make a case for it:

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/value

Like, if the only reason to have a separate tag is to describe _an implementation detail that the caller shouldn't even know about_ (in this example, they're explaining the private backing field for the property) then why would this be a part of generated API documentation?

eli-darkly: Also worth noting: when doc generation is turned on, the compiler spits out warnings if you omit a `<summary>` or `<param>` or `<returns>`, but it doesn't bother if you omit `<value>`.

src/LaunchDarkly.EventSource/ByteArrayLineScanner.cs
gwhelanLD: This comment ended up being more stream of consciousness than a true question, so I would consider it resolved anyway. I'll clarify what I was originally unsure about however.

My question was more why we only check `_startPos == _count`, on line 149, rather than doing a `_startPos < _count` and performing this same `BlockCopy` to shift the bytes and make room for more input. In the case that the provided bytes terminate in a line ending, the check on 149 is sufficient, but will leave already consumed input taking space in the buffer in other cases.

My confusion was just about why we would bother with the `BlockCopy` in one case but not the other, but I believe the statement after my previous question explains the difference by considering that the network data being given to the SSE stream will normally be line-chunked anyway.

eli-darkly: No, the assumption was _not_ that the stream will normally be line-chunked. That may be true, but it's definitely not why I wrote the code this way. Let me take another look and see if I can understand better what you're getting at.

eli-darkly: Oh wait, I see what you're saying. There is a different rationale, but it's not self-evident and I should have made it clear in the comments.

`ScanToEndOfLine`, whenever possible, returns a span that points to _the original bytes_, not a copy of them. This is because it's likely that we'll end up copying or appending the bytes to another buffer further down the line, so there's no point in allocating a new temporary buffer unnecessarily. For instance, if the content of the buffer looks like this—

```
[0] 'a'
[1] 'b'
[2] \n
[3] 'c'
[4] 'd'
[5] \e
```

—the first call to the method will return a span that points to bytes 0 through 1, and the second will return a span that points to bytes 3 through 4, in the buffer itself. So, we don't want to shift the data over immediately after we scan the first line, because that would overwrite the bytes that the caller is about to be looking at.

The case where `_partialLine` exists is different, because there we have already allocated a separate buffer— we had to, because the data wouldn't fit in the main fixed-size one— and so we can safely return a reference to that data while relocating the bytes in the main buffer.

gwhelanLD: Oh right, I had known those were just views but I wasn't thinking about it when reading this code. Thanks for the explanation, that certainly explains why we can do the `BlockCopy` in the first case but not the second.

src/LaunchDarkly.EventSource/ByteArrayLineScanner.cs
```diff
@@ -0,0 +1,179 @@
...
+            }
+            else
+            {
+                lineOut = new Utf8ByteSpan(_buffer, startedAt, pos - startedAt);
+                if (_startPos == _count)
+                {
+                    // If we've scanned all the data in the buffer, reset _startPos and _count to indicate
+                    // that the entire buffer is available for the next read. It's safe for us to do this
+                    // before the caller has looked at lineOut, because we're not actually modifying any
+                    // bytes in the buffer. It's the caller's not responsibility not to modify the buffer
```
gwhelanLD, line 171: Minor wording typo:
```suggestion
                    // bytes in the buffer. It's the caller's responsibility not to modify the buffer
```

gwhelanLD (approved):
One minor wording issue on the new comments, otherwise looks good to me.

src/LaunchDarkly.EventSource/ByteArrayLineScanner.cs
eli-darkly: 🤦

PR #64, opened 2021-01-16
API cleanup 1: simpler configuration semantics, hide impl details, code cleanup

eli-darkly:
Some improvements for the next major version, some of which are backward-incompatible.

* `Configuration` can only be created with `ConfigurationBuilder`.
* New `EventSource(Uri)` constructor is a shortcut for if you don't need any other configuration.
* Previously, `ConfigurationBuilder` methods could throw exceptions if a value was out of range or inconsistent with other values. Now, they no longer throw any exceptions, unless you didn't provide a URI. Out-of-range values are simply pinned to the range; if two properties are incompatible (like `HttpClient` and `HttpMessageHandler`), there's a well-defined precedence as to which one will be used. This is in line with the LaunchDarkly SDK's general philosophy of making configuration builders safe unless there's absolutely no way to make a valid configuration out of the inputs.
* `MessageHandler` config property is renamed to `HttpMessageHandler` since that's what it really is ("message" has other meanings in this library).
* Removed the unnecessary type `Configuration.HttpContentFactory` which is really just a `Func<HttpContent>`.
* If you pass a dictionary of headers into the `ConfigurationBuilder`, it gets copied rather stored by reference, for safety.
* Fixed many malformed, missing, or unclearly written doc comments. We weren't actually generating XML documentation before, so we weren't getting build warnings for comment problems. I've also started standardizing the style to be more consistent with the LaunchDarkly SDK API documentation: for instance, parameter descriptions which were formerly sentences ("Description of this parameter.") are now phrases ("description of this parameter").
* Alphabetized configuration methods.
* Removed deprecated things.


src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -2,29 +2,48 @@
...
+        /// The maximum value for <see cref="ConfigurationBuilder.DelayRetryDuration(TimeSpan)"/>:
+        /// 30 seconds.
+        /// </summary>
+        public static readonly TimeSpan MaximumRetryDuration = TimeSpan.FromSeconds(30);
 
-        #region Constants
+        /// <summary>
+        /// The default value for <see cref="ConfigurationBuilder.ConnectionTimeout(TimeSpan)"/>:
+        /// 10 seconds.
+        /// one second.
```
gwhelanLD, line 33: Extraneous line.

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -2,29 +2,48 @@
...
+        /// </summary>
+        public static readonly TimeSpan DefaultDelayRetryDuration = TimeSpan.FromSeconds(1);
 
-        #endregion
+        /// <summary>
+        /// The maximum value for <see cref="ConfigurationBuilder.DelayRetryDuration(TimeSpan)"/>:
+        /// 30 seconds.
+        /// </summary>
+        public static readonly TimeSpan MaximumRetryDuration = TimeSpan.FromSeconds(30);
```
gwhelanLD, line 28: It seems a bit inconsistent to call this `MaximumRetryDuration` rather than `MaximumDelayRetryDuration`. Do you think this is something worth changing?

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -129,114 +121,59 @@ public sealed class Configuration
...
+        /// <seealso cref="ConfigurationBuilder.RequestHeader(string, string)"/>
+        /// <seealso cref="ConfigurationBuilder.RequestHeaders(IDictionary{string, string})"/>
+        public IDictionary<string, string> RequestHeaders { get; }
+
+        /// <summary>
+        /// Gets the <see cref="System.Uri"/> used when connecting to an EventSource API.
         /// </summary>
         /// <value>
-        /// The maximum duration of the retry.
+        /// The <see cref="System.Uri"/>.
```
gwhelanLD, line 149: This seems to be the only property here with a `<value>` tag. As I mentioned in another PR, I'm not entirely sure how important these are for the code doc views (generated HTML or in IDE), but maybe we could be more consistent with these.

eli-darkly: We had a bunch of them, I took most of them out, I missed this one and will take it out now.

src/LaunchDarkly.EventSource/Configuration.cs
eli-darkly: I'd been trying to avoid making breaking changes that were _only_ for better naming, but I agree that these are weirdly named— both of them. "DelayRetryDuration" is weird because it looks like it should mean "duration of the retry of the delay", when it's really "duration of the delay before the retry". And "MaximumRetryDuration" is wrong because it's not a duration of a retry.

eli-darkly: And I think "Duration" is redundant in both cases because a delay that is represented by a TimeSpan is, by definition, a duration.

gwhelanLD: Huh, I was actually confused about what these were doing, probably partially because of the names. 

I was thinking this was the default value of a configurable property that would control the maximum delay time of the backoff. While it is the maximum backoff delay time, it is not a configurable property, and also functions as the maximum value of the initial retry delay.

Maybe this isn't the right PR to bring this up in, but this behavior doesn't make a lot of sense to me. I don't see the value in limiting the maximum value of the initial backoff delay. It would make more sense to me to just allow configuring the maximum delay of the backoff.

If this doesn't makes sense right now or I'm missing something, then maybe we could just include a line in the documentation comment for `MaximumRetryDuration` to mention that it is also the maximum delay after backoffs as well.

eli-darkly: > I don't see the value in limiting the maximum value of the initial backoff delay. It would make more sense to me to just allow configuring the maximum delay of the backoff.

That's reasonable, and it corresponds to how some of our other EventSource implementations are configured.

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -99,23 +109,10 @@ public sealed class Configuration
...
         /// <summary>
-        /// The request headers to be sent with each EventSource HTTP request.
-        /// </summary>
-        /// <seealso cref="ConfigurationBuilder.RequestHeader(string, string)"/>
-        /// <seealso cref="ConfigurationBuilder.RequestHeaders(IDictionary{string, string})"/>
-        public IDictionary<string, string> RequestHeaders { get; }
-
-        /// <summary>
-        /// The HttpMessageHandler that will be used for the HTTP client, or null for the default handler.
+        /// Sets the maximum amount of time to wait before attempting to reconnect.
```
gwhelanLD, line 112: Last nitpick:
```suggestion
        /// The maximum amount of time to wait before attempting to reconnect.
```

gwhelanLD (approved):
One last nitpick on a doc comment, otherwise looks good to me.

PR #65, opened 2021-01-16
API cleanup 2: move the event name into the MessageEvent type

eli-darkly:
Previously, `MessageReceivedEventArgs`, the type that's passed to a `MessageReceived` event handler, had two properties: the event name, and a `MessageEvent` struct containing everything else (Data, etc.). This somewhat idiosyncratic design choice meant that you would often have to pass two parameters around when doing anything with a received event, since the event data is unlikely to be meaningful without the event name or vice versa. Since they are conceptually all properties of a single event in the SSE spec, I've put the name into `MessageEvent`.

`MessageReceivedEventArgs` is now just a wrapper for that object, but I've retained it because 1. it's a .NET convention to have an event parameters class like this that descends from `EventArgs` even if it only has one property, and 2. there's less adjustment for existing application code. I kept `MessageReceivedEventArgs.EventName` as a shortcut for getting the name out of the event.


src/LaunchDarkly.EventSource/MessageReceivedEventArgs.cs
```diff
@@ -3,36 +3,29 @@
...
+        /// SSE stream.
         /// </summary>
-        /// <value>
-        /// The <see cref="MessageEvent"/> data recieved by the Server Sent Event.
-        /// </value>
         public MessageEvent Message { get; }
 
         /// <summary>
-        /// Gets the name of the event type received by the Server Sent Event.
+        /// Shortcut for getting the <see cref="MessageEvent.Name"/> property of th eevent.
```
gwhelanLD, line 18: typo: `th eevent`.

gwhelanLD (approved):
One typo, otherwise looks good to me.

PR #66, opened 2021-01-29
general unit test improvements

eli-darkly:
The tests in this project were based on some pretty old code that we had added to here and there without much attention to DRY or consistent style. I've reorganized them into clearer functional areas (e.g. `EventSourceReconnectingTest`) and streamlined many of them by using a simple event-receiving fixture. Also, logs from both the EventSource and the fixture are now sent to the Xunit test output buffer so you can see them if a test fails.

I also made a couple of non-test changes based on some slightly indeterministic behavior I spotted in the tests - see PR comments.


src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -187,8 +187,6 @@ public async Task StartAsync()
                     // If the user called Close(), ReadyState = Shutdown, so errors are irrelevant.
                     if (ReadyState != ReadyState.Shutdown)
                     {
-                        Close(ReadyState.Closed);
-
```
eli-darkly, line 191: Changed this so the transition to the closed state happens after it's called OnError. May not matter for real applications, but it makes more sense in the tests to see (error, closed) rather than (closed, error).

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -154,7 +154,7 @@ private Encoding DetectEncoding(HttpResponseMessage response)
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                string line = await DoTaskWithReadTimeout(() => reader.ReadLineAsync(), needManualTimeout);
+                string line = await DoTaskWithReadTimeout(reader.ReadLineAsync, needManualTimeout);
```
eli-darkly, line 157: This is just syntactic simplification - the method reference by itself does the same thing as a lambda that calls the method.

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -208,7 +208,7 @@ private async Task<T> DoTaskWithReadTimeout<T>(Func<Task<T>> task, bool needManu
             }
 
             var timeoutCancellation = new CancellationTokenSource();
-            var readTimeoutTask = Task.Delay(_configuration.ReadTimeout, timeoutCancellation.Token);
+            var readTimeoutTask = Task.Run(() => Task.Delay(_configuration.ReadTimeout, timeoutCancellation.Token));
```
eli-darkly, line 211: I'm not 100% sure but it does seem as if using `Task.Run`, rather than just creating a task with a method like `Delay`, causes scheduling to be more predictable.

test/LaunchDarkly.EventSource.Tests/ConfigurationBuilderTest.cs
```diff
@@ -1,246 +1,246 @@
...
-        [Fact]
-        public void BuilderSetsUri()
-        {
-            var b = Configuration.Builder(uri);
-            Assert.Equal(uri, b.Build().Uri);
-        }
-
-        [Fact]
-        public void BuilderRejectsNullUri()
-        {
```
eli-darkly, line 23: The changes here aren't real - the file just had inconsistent linefeeds in it.

test/LaunchDarkly.EventSource.Tests/EventParserTest.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace LaunchDarkly.EventSource.Tests
 {
-    public class EventParserTests
+    public class EventParserTest
```
eli-darkly, line 6: Trivial, but our usual naming style is to call a test class `___Test` instead of `___Tests`, since conceptually Xunit instantiates this class once for each test.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -0,0 +1,196 @@
...
+using EmbedIO.Routing;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace LaunchDarkly.EventSource.Tests
+{
+    // For most of the EventSource tests, it's adequate to use a fake HttpMessageHandler that provides
+    // the desired response data without really doing HTTP. But we do want to be sure that our I/O logic
+    // works as intended with a real HTTP client, so this class does some basic EventSource tests
+    // against an embedded HTTP server.
```
eli-darkly, line 19: The embedded server library I'm using here is `EmbedIO`. I've used `WireMock.Net` for similar things in the past, but I've had some issues with that one due to the way it tries to use various Microsoft server frameworks under the covers depending on what platform you're on, whereas `EmbedIO` is a self-contained implementation. I didn't end up having to use it for much, since in the other test classes I think a fake message handler is adequate, but I'll keep it in mind for any future tests like this.

test/LaunchDarkly.EventSource.Tests/EventSourceHttpBehaviorTest.cs
```diff
@@ -0,0 +1,193 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace LaunchDarkly.EventSource.Tests
+{
+    public class EventSourceHttpBehaviorTest : BaseTest
```
eli-darkly, line 13: These tests aren't new, they're just moved out of the former `EventSourceTests` and DRYd up a bit.

src/LaunchDarkly.EventSource/EventSourceService.cs
eli-darkly: That turned out to be not quite the issue, but I'll explain in the next PR where (I think) I greatly improved how timeouts work.

test/LaunchDarkly.EventSource.Tests/EventSourceReconnectingTest.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        [Fact]
+        public void RetryDelayDurationsShouldIncrease()
+        {
+            var handler = new StubMessageHandler();
+
+            var nAttempts = 3;
+            for (var i = 0; i < nAttempts; i++)
+            {
+                handler.QueueResponse(StubResponse.StartStream(StreamAction.CloseStreamAbnormally()));//StubResponse.WithIOError());
```
gwhelanLD, line 81: Looks like a bit of extraneous commented code.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
```diff
@@ -1,628 +0,0 @@
...
-            var request = handler.GetRequests().First();
-            Assert.True(headers.All(
-                    item =>
-                        request.Headers.Contains(item.Key) &&
-                        request.Headers.GetValues(item.Key).Contains(item.Value)
-            ));
-        }
-
-        [Fact]
-        public async Task Given_content_type_not_equal_to_eventstream_when_the_http_response_is_received_then_error_event_should_occur()
```
gwhelanLD, line 346: We had some serious sentence oriented programming going on in this file.

gwhelanLD (approved):
Looks good, found one tiny bit of commented extraneous code to be removed.

test/LaunchDarkly.EventSource.Tests/EventSourceTests.cs
eli-darkly: Yeah, I try not to go too far in the opposite direction, it's possible of course to have overly cryptic short test names, but I do feel like (totally aside from the "complete grammatical sentence" vs. "abbreviated phrase" issue) it's overkill to try to fully describe both the test preconditions and the test assertions in the name of the test. The main purposes of the name of the test are that if the test fails, you can get an idea of what kind of functionality may have failed, and you can easily locate the relevant test code. The test code itself usually shows you what you need to know about the details of the preconditions/assertions (like, we set the content type to X, or we detected event X) or if it can't be made self-evident then code comments serve the same purpose. And concise method names (IMO) actually make it a bit easier to see what kind of tests we have, as a larger proportion of the method name is made up of words that uniquely identify that test.

Anyway, the style differences in this repo are just due to the first version of it having been written fairly quickly a few years ago by a contractor, at a time when we didn't have much of a .NET codebase to compare to.

PR #67, opened 2021-01-30
add Restart method and improve timeout/cancellation logic

eli-darkly:
[ch99689](https://app.clubhouse.io/launchdarkly/story/99689/add-restart-method-to-net-eventsource) explains the need for a Restart method.

In order to implement that, I had to update the I/O logic to make cancelling a pending read task from the outside (that is, with a CancellationToken, rather than having the task exit because it got an I/O error) work correctly... which led to a bunch of other changes. I think these changes are helpful both in terms of efficient and reliable behavior in real usage, and also for making the tests more deterministic. See PR comments for details.


src/LaunchDarkly.EventSource/AsyncHelpers.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Threading.Tasks;
+
+namespace LaunchDarkly.EventSource
+{
+    internal static class AsyncHelpers
+    {
+        // General-purpose timeout logic that uses a timed CancellationToken instead of
+        // Task.Delay. The cancellationToken that is passed in is for explicitly cancelling
+        // the operation from elsewhere; DoWithTimeout wraps this to create a token that can
+        // also be cancelled by a timeout, and passes that token to taskFn.
```
eli-darkly, line 12: This replaces the previous logic where, in order to do an async read with a timeout, we would create an additional task that _is_ the timeout (with `Task.Delay`) and then use `Task.WhenAny` to see which one of those two tasks (the read and the timeout) finished first.

.NET has a better way to introduce timeouts into async tasks: create a `CancellationToken` with a `TimeSpan` parameter, which causes it to fire a cancellation after that amount of time. We already have another `CancellationToken` that carries the "we are explicitly cancelling because they called either Close or Restart" signal, so we wire them together with `CreateLinkedTokenSource`. Now we have a single cancel signal, and if it goes off, we can tell whether it was a timeout by checking whether the original "explicitly cancel" signal was used or not.

This is still slightly trickier than it sounds because .NET's support for cancellation tokens is inconsistent - see below.

src/LaunchDarkly.EventSource/AsyncHelpers.cs
```diff
@@ -0,0 +1,91 @@
...
+                    throw;
+                }
+                throw new ReadTimeoutException();
+            }
+        }
+
+        // Take a task that can't be terminated early with a CancellationToken, and wrap it in a task
+        // that can. If you cancel the wrapped task, the inner task will still continue, but we're
+        // assuming that in these situations we'll eventually be closing the underlying socket which
+        // will terminate any pending I/O.
```
eli-darkly, line 42: Here's the other thing we need for our new improved timeout logic. While the idea of the cancellation token is very nice, and it works well for things like HTTP requests, some of the basic .NET I/O methods either don't have a cancellation token parameter (like `StreamReader.ReadAsync`), or do have one but it doesn't actually work (like `Stream.ReadAsync`— that is, the base class `Stream` declares a method with that parameter, but some of the stream subclasses don't implement it, and unfortunately the one for reading a socket is one of those).

So, we have to use the Task API in a somewhat confusing way to make it so we really can interrupt these operations... or at least stop waiting on them. See code comments below.

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -187,7 +187,11 @@ public async Task StartAsync()
                     if (ReadyState != ReadyState.Shutdown)
                     {
                         Exception realException = e;
-                        if (e is OperationCanceledException oe)
+                        if (realException is AggregateException ae && ae.InnerExceptions.Count == 1)
```
eli-darkly, line 190: A somewhat frustrating feature of the Task API is that if you've done any kind of wrapping or decorating of a task, exceptions that are thrown out of it might appear in their original form, or they might be wrapped in an AggregateException.

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -99,30 +102,26 @@ CancellationToken cancellationToken
 
                 using (var stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false))
                 {
-                    bool needManualTimeout = false;
-                    try
-                    {
-                        // Some Stream subclasses on some platforms have their own read timeout implementation; others
-                        // don't. The only way to find out is to try to set it and see if you get an exception. If we
-                        // get an exception, then ReadTimeout won't work and we'll need to implement our own timeout.
-                        stream.ReadTimeout = (int)_configuration.ReadTimeout.TotalMilliseconds;
```
eli-darkly, line 108: This logic turned out to be useless because the Stream subclass that you get for a network socket, for whatever reason, is not one of the ones that supports a read timeout natively.

src/LaunchDarkly.EventSource/EventSourceStreamReader.cs
```diff
@@ -1,49 +0,0 @@
...
-        private StreamReader _streamReader;
-
-        public EventSourceStreamReader(Stream stream, Encoding encoding)
-        {
-            _streamReader = new StreamReader(stream, encoding);
-        }
-
-        public virtual async Task<string> ReadLineAsync()
-        {
-            await Task.Yield();
```
eli-darkly, line 25: I'm really not sure why we created a whole decorator class just to add this one line before each read. If `Task.Yield()` is helpful, we can simply add it before calling `ReadLineAsync()` [here](https://github.com/launchdarkly/dotnet-eventsource/pull/67/files#diff-0e94183352c89bfec8bf05e34839bedc8a4c7ef1adf4348ae42561f87caca04aR156). But I'm not convinced that it is helpful and so I haven't added it there. I think the intention was to in effect make the stream reading task lower-priority, by having it offer up some extra time to other tasks before each of its read operations, but I'm not sure that that's really how .NET task scheduling works— the scheduler should not allow this task to starve other tasks anyway, there is an implicit yield every time you do async I/O— and in any case unless you're getting a very rapid constant flow of stream data, which is not a typical use case for SSE, these reads are not being done very often.

test/LaunchDarkly.EventSource.Tests/BaseTest.cs
```diff
@@ -39,6 +40,14 @@ public BaseTest()
...
             _testLogging = _logCapture;
             _testLogger = _testLogging.Logger("");
+
+            // The following line prevents intermittent test failures that can happen due to the low
+            // default setting of ThreadPool.SetMinThreads causing new worker tasks to be severely
+            // throttled: http://joeduffyblog.com/2006/07/08/clr-thread-pool-injection-stuttering-problems/
+            // This makes it difficult to test things such as timeouts. We believe it not to be a real
+            // issue in non-test scenarios, since the tests are starting and stopping an unusually large
+            // number of async tasks in a way that regular use of EventSource would not do.
+            ThreadPool.SetMinThreads(100, 100);
```
eli-darkly, line 50: This had bitten me before when I was debugging the `dotnet-server-sdk` tests and seeing that a lot of tests either took a couple seconds just to start up, or had an inexplicable delay of a couple seconds prior to some trivial async operation (you can see this locally if you just remove this line and run the tests: even if you don't get spurious failures for timeout-related stuff, you'll see that they run much, much slower).

What I found was that, at least on some runtimes, the thread pool size in the Xunit environment is _two_. And if at any time you try to start a Task of any kind (even if it's just a `Task.Delay` that doesn't in itself _do_ anything) when there are already two threads in use from the pool, the scheduler can decide to throttle you by simply pausing for a couple seconds, even though not much else is going on.

That's something that could theoretically be an issue in a real application too, depending on how things are configured, but that wouldn't only affect EventSource. I don't think we are creating a particularly large number of tasks in this code compared to any other async library, it's just that we're running dozens of tests within a short period of time, most of which start a new async EventSource client which might not completely finish shutting down before the next test runs.

src/LaunchDarkly.EventSource/AsyncHelpers.cs
eli-darkly: And I think this is still better than our earlier logic that used `WhenAny`, because that was always creating two _simultaneously active_ tasks every time we wanted to do a read. Here, using `ContinueWith` does create additional tasks, but they are chained— `cancellableTask` isn't taking up any room in the scheduler until `originalTask` finishes.

gwhelanLD (approved):

eli-darkly:
@gwhelanLD - I did a little streamlining of the AsyncHelpers logic along the lines that you suggested, and added some tests that make me feel better that it all works as intended. I removed the "doesn't cause an unobserved exception" test that was previously in EventSourceStreamReadingTest, because it was really just an indirect test of the AsyncHelpers timeout logic that is now being tested more directly.

gwhelanLD (approved):

PR #68, opened 2021-02-02
misc project cleanup/metadata improvements

eli-darkly:
See: https://github.com/launchdarkly/dotnet-logging/pull/4


src/LaunchDarkly.EventSource/LaunchDarkly.EventSource.csproj
```diff
@@ -1,23 +1,24 @@
...
-    <Description>LaunchDarkly .NET Event Source library for Server Sent Events</Description>
+    <Description>LaunchDarkly Server-Sent Events client</Description>
     <PackageProjectUrl>https://github.com/launchdarkly/dotnet-eventsource</PackageProjectUrl>
     <RepositoryUrl>https://github.com/launchdarkly/dotnet-eventsource</RepositoryUrl>
-    <SignAssembly>true</SignAssembly>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LaunchDarkly.Logging" Version="1.0.0" />
+    <PackageReference Include="LaunchDarkly.Logging" Version="[1.0.0,2.0.0)" />
```
eli-darkly, line 21: I've been moving toward using dependency version ranges instead of single versions, for dependencies like this that are referencing an API whose actual version is expected to be overridden by the application or by a further-up library (as opposed to something like the dependency from `LaunchDarkly.ServerSdk` to `LaunchDarkly.InternalSdk`, where the latter is conceptually an integral part of the former and an exact version dependency is appropriate). I'm not sure it makes a big difference in practice, as the default behavior (in .NET Core at least) is already to allow lower versions of a transitive dependency to be overridden by higher ones, but I think this makes the intention clearer.

bwoskow-ld (approved):

PR #70, opened 2021-02-20
make CI tests work in Windows

eli-darkly:
There were two reasons why we didn't have CI tests for this library running in Windows till now:

1. Originally, CircleCI didn't support it.
2. Then, we tried and found out that the WireMock.Net library we were using for an embedded HTTP server had some compatibility problems in .NET Framework.

Those are no longer relevant, but I found a few more test issues - see PR comments.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,10 @@ workflows:
           name: .NET Core 3.1
           docker-image: mcr.microsoft.com/dotnet/core/sdk:3.1-focal
           test-target-framework: netcoreapp3.1
+      - test-netcore:
+          name: .NET 5.0
+          docker-image: mcr.microsoft.com/dotnet/sdk:5.0-focal
+          test-target-framework: net5.0
       - test-windows-netframework-4-6-1:
           name: .NET Framework 4.6.1
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,50 @@
 version: 2.1
 
+orbs:
+  win: circleci/windows@1.0.0
+
 workflows:
   version: 2
   test:
     jobs:
-      - test-netcore-2-1
+      - test-netcore:
+          name: .NET Core 2.1
+          docker-image: mcr.microsoft.com/dotnet/core/sdk:2.1-focal
+          test-target-framework: netcoreapp2.1
+      - test-netcore:
+          name: .NET Core 3.1
+          docker-image: mcr.microsoft.com/dotnet/core/sdk:3.1-focal
+          test-target-framework: netcoreapp3.1
+      - test-windows-netframework-4-6-1:
+          name: .NET Framework 4.6.1
 
 jobs:
-  test-netcore-2-1:
+  test-netcore:
+    parameters:
+      docker-image:
+        type: string
+      test-target-framework:
+        type: string
     docker:
-      - image: mcr.microsoft.com/dotnet/core/sdk:2.1-focal
+      - image: <<parameters.docker-image>>
+    environment:
+      ASPNETCORE_SUPPRESSSTATUSMESSAGES: true
+      TESTFRAMEWORK: <<parameters.test-target-framework>>
     steps:
       - run:
           name: install packages
           command: apt-get -q update && apt-get install -qy awscli
       - checkout
-      - run: dotnet restore
       - run: dotnet build src/LaunchDarkly.EventSource -f netstandard2.0
-      - run: dotnet test test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj -f netcoreapp2.1
+      - run: dotnet test test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj -f <<parameters.test-target-framework>>
+
+  test-windows-netframework-4-6-1:
+    executor:
+      name: win/vs2019
+      shell: powershell.exe
+    environment:
+      TESTFRAMEWORK: net461
+    steps:
+      - checkout
+      - run: dotnet build src/LaunchDarkly.EventSource -f net452
+      - run: dotnet test test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj -f net461
--- a/test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
+++ b/test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
@@ -29,100 +29,78 @@ namespace LaunchDarkly.EventSource.Tests
             // parsed correctly regardless of how the chunks line up with the events.
 
             var eventData = new List<string>();
-            var chunks = new List<string>();
             for (var i = 0; i < 200; i++)
             {
                 eventData.Add(string.Format("data{0}", i) + new string('x', i % 7));
             }
             var allBody = string.Concat(eventData.Select(data => "data:" + data + "\n\n"));
-            for (var pos = 0; ;)
+            var allEventsReceived = new EventWaitHandle(false, EventResetMode.ManualReset);
+
+            IEnumerable<string> DoChunks()
             {
-                int i = chunks.Count;
-                int chunkSize = i % 20 + 1;
-                if (pos + chunkSize >= allBody.Length)
+                var i = 0;
+                for (var pos = 0; ;)
                 {
-                    chunks.Add(allBody.Substring(pos));
-                    break;
+                    int chunkSize = i % 20 + 1;
+                    if (pos + chunkSize >= allBody.Length)
+                    {
+                        yield return allBody.Substring(pos);
+                        break;
+                    }
+                    yield return allBody.Substring(pos, chunkSize);
+                    pos += chunkSize;
+                    i++;
                 }
-                chunks.Add(allBody.Substring(pos, chunkSize));
-                pos += chunkSize;
+                allEventsReceived.WaitOne();
             }
 
-            using (var server = StartWebServerOnAvailablePort(out var uri, ctx =>
+            try
             {
-                ctx.Response.ContentType = "text/event-stream";
-                ctx.Response.SendChunked = true;
-                var stream = ctx.Response.OutputStream;
-                foreach (var chunk in chunks)
+                using (var server = StartWebServerOnAvailablePort(out var uri,
+                    RespondWithChunks("text/event-stream", DoChunks)))
                 {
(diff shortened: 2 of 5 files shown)


test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -29,100 +29,78 @@ public void ReceiveEventStreamInChunks()
...
 
             var eventData = new List<string>();
-            var chunks = new List<string>();
             for (var i = 0; i < 200; i++)
             {
                 eventData.Add(string.Format("data{0}", i) + new string('x', i % 7));
             }
             var allBody = string.Concat(eventData.Select(data => "data:" + data + "\n\n"));
-            for (var pos = 0; ;)
+            var allEventsReceived = new EventWaitHandle(false, EventResetMode.ManualReset);
```
eli-darkly, line 37: Here I added some synchronization (see line 54 and line 80) to ensure that the HTTP handler won't exit before the client side of the test has caught up to it.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -29,100 +29,78 @@ public void ReceiveEventStreamInChunks()
...
-                    break;
+                    int chunkSize = i % 20 + 1;
+                    if (pos + chunkSize >= allBody.Length)
+                    {
+                        yield return allBody.Substring(pos);
+                        break;
+                    }
+                    yield return allBody.Substring(pos, chunkSize);
+                    pos += chunkSize;
+                    i++;
```
eli-darkly, line 52: The logic here for making the chunked hasn't really changed, it's just that I switched from using an array of strings to having a generator function that `yield return`s the strings one at a time - not a big difference, just seemed a little cleaner.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -29,100 +29,78 @@ public void ReceiveEventStreamInChunks()
...
+                        expectedActions.Add(EventSink.MessageReceivedAction(new MessageEvent(MessageEvent.DefaultName, data, uri)));
+                    }
+
+                    var config = Configuration.Builder(uri).LogAdapter(_testLogging).Build();
+                    using (var es = new EventSource(config))
+                    {
+                        var sink = new EventSink(es);
+                        _ = es.StartAsync();
+                        sink.ExpectActions(expectedActions.ToArray());
+                    }
```
eli-darkly, line 75: The logic in this block hasn't changed, it just got indented differently.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -29,100 +29,78 @@ public void ReceiveEventStreamInChunks()
...
-                }
+                allEventsReceived.Set();
             }
         }
 
-        [Fact]
-        public void ReadTimeoutIsDetectedInRawUtf8ReadingMode()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
```
eli-darkly, line 86: I replaced the two very similar `ReadTimeoutIsDetected...` test methods with a single parameterized one.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -137,20 +115,16 @@ public void ReadTimeoutIsDetectedInRawUtf8ReadingMode()
...
         private static WebServer StartWebServerOnAvailablePort(out Uri serverUri, Action<IHttpContext> handler)
         {
             var module = new SimpleModule(handler);
 
             for (int port = 10000; ; port++)
             {
-                var server = new WebServer(port).WithModule(module);
+                var options = new WebServerOptions()
+                    .WithUrlPrefix($"http://*:{port}")
+                    .WithMode(HttpListenerMode.EmbedIO);
```
eli-darkly, line 126: EmbedIO can run in two modes: the EmbedIO mode, where it always uses its own web server implementation, or a mode where it uses some built-in Windows stuff if that is available. Using the former makes the behavior more predictable cross-platform.

test/LaunchDarkly.EventSource.Tests/EventSourceReconnectingTest.cs
```diff
@@ -82,7 +82,7 @@ public void RetryDelayDurationsShouldIncrease()
             {
                 handler.QueueResponse(StubResponse.StartStream(
                     StreamAction.Write(":hi\n"),
-                    StreamAction.CloseStreamAbnormally()));
```
eli-darkly, line 85: I'm not sure what I intended the difference between CloseStream and CloseStreamAbnormally to be, but it didn't really contribute anything to this test.

test/LaunchDarkly.EventSource.Tests/LaunchDarkly.EventSource.Tests.csproj
```diff
@@ -1,13 +1,19 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
+    <!-- The TESTFRAMEWORK variable allows us to override the target frameworks with a
+         single framework that we are testing; this allows us to test with older SDK
+         versions that would error out if they saw any newer target frameworks listed
+         here, even if we weren't running those. -->
```
eli-darkly, line 6: I've done this kind of thing in most of our .NET libraries, ever since we started running CI tests for newer frameworks like .NET 5.0. The issue is that when we run the CI job for, say, .NET Core 2.1, we want to actually use the .NET Core 2.1 SDK— but that SDK will complain if it even sees `net5.0` listed here in the project file.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -57,140 +58,141 @@ public HttpRequestMessage AwaitRequest()
         }
     }
 
-    public abstract class StubResponse
```
eli-darkly, line 60: Most of this is just linefeed changes. I've tried to normalize all the linefeeds with `.gitattributes` but there were some files that hadn't been updated.

test/LaunchDarkly.EventSource.Tests/Stubs/StubHandler.cs
```diff
@@ -57,140 +58,141 @@ public HttpRequestMessage AwaitRequest()
...
+            _encoding = encoding ?? Encoding.UTF8;
+            _specifiedEncoding = encoding != null;
+        }
+
+        override public HttpResponseMessage MakeResponse(CancellationToken cancellationToken)
+        {
+            var httpResponse = new HttpResponseMessage(HttpStatusCode.OK);
+            var pipe = new Pipe();
+            var readStream = pipe.Reader.AsStream(true);
+            var content = new StreamContent(readStream);
```
eli-darkly, line 134: This part is the most significant change. For the tests that don't need a real embedded HTTP server, we use this fake HTTP handler that pipes data into the stream. I was using this other pipe mechanism (see old lines 131-132) which is really intended for IPC, and it seems that it doesn't behave very predictably when you use it within a single process: the behavior I saw was that in Windows, reading from `streamRead` when there wasn't yet any more data waiting would not block as it should, but would return "0 bytes" as if the stream was actually finished. And that would cause EventSource to think the server had closed the request, which would mess up tests in various ways.

Anyway, using `Pipe` from `System.IO.Pipelines` makes this behave as it should.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -29,100 +29,78 @@ public void ReceiveEventStreamInChunks()
...
         {
             TimeSpan readTimeout = TimeSpan.FromMilliseconds(200);
-            using (var server = StartWebServerOnAvailablePort(out var uri, ctx =>
+            IEnumerable<string> DoChunks()
             {
-                ctx.Response.ContentType = "text/event-stream";
-                ctx.Response.SendChunked = true;
-                var stream = ctx.Response.OutputStream;
-                WriteChunk(stream, "data: event1\n\ndata: e");
+                yield return "";
```
eli-darkly, line 92: This is a bit weird but I found that if I did not write a zero-length chunk at the start of the stream like this, `Stream.ReadAsync` would not pick up the next non-zero-length chunk right away but only when the next one had been written— as if the first one had not been flushed— which would mess up any tests that involved timeouts. This would only happen in Windows, and I couldn't find any documented reason for it; there doesn't seem to be any such requirement for chunked HTTP responses, in fact a zero-length chunk isn't even supposed to be a thing in chunked HTTP. But it does consistently make a difference. I'm not too concerned about the mysteriousness of this because we are verifying in many other tests that the client code is able to read chunked streams correctly; it's only this fake server whose behavior is weird, and that's not part of the library.

.circleci/config.yml
```diff
@@ -1,20 +1,54 @@
...
+      - test-netcore:
+          name: .NET Core 2.1
+          docker-image: mcr.microsoft.com/dotnet/core/sdk:2.1-focal
+          test-target-framework: netcoreapp2.1
+      - test-netcore:
+          name: .NET Core 3.1
+          docker-image: mcr.microsoft.com/dotnet/core/sdk:3.1-focal
+          test-target-framework: netcoreapp3.1
+      - test-netcore:
+          name: .NET 5.0
```
eli-darkly, line 19: The library itself will be the same .NET Standard 2.0 build in all three of these Linux-based jobs, but since there can be unexpected differences in behavior of the standard runtime library across different platform versions, I've been standardizing the CI jobs for all of our .NET libraries to test them in each of the major Core releases.

.circleci/config.yml
```diff
@@ -1,20 +1,54 @@
...
+          docker-image: mcr.microsoft.com/dotnet/sdk:5.0-focal
+          test-target-framework: net5.0
+      - test-windows-netframework-4-6-1:
+          name: .NET Framework 4.6.1
```
eli-darkly, line 23: The lower bound of our .NET Framework compatibility for this library is actually 4.5.2. The problem is that we do need an embedded HTTP server for some of our tests; WireMock.Net theoretically works in 4.5.2, but in practice it's been pretty unreliable, and EmbedIO isn't available for 4.5.2.

gwhelanLD (approved):

PR #71, opened 2021-02-22
clean up error messages

eli-darkly:
This library's error messages, which date back to a pretty old first version from before we had any style conventions at all, have some idiosyncratic choices:

* Referring to the server endpoint as the "EventSource API" (`A timeout occurred while reading the request from the remote EventSource API`)— that's incorrect, the EventSource API would either be this library's own EventSource class or the _browser JavaScript API_ of the same name; the protocol used by the server is called Server-Sent Events.
* Random capitalization (`Remote EventSource API returned Http Status Code 204.`), (`Http Response contained no content.`)
* Sometimes ending with a period, sometimes not
* The programmatic names for the messages(*) are using underscore delimiters, inconsistent with .NET symbol naming conventions.

(* We're using Visual Studio's resource file mechanism for internationalizable messages, even though we've only provided English-language text; this causes each string to be defined as a named constant in the Resources class.)

### DIFF
--- a/src/LaunchDarkly.EventSource/EventSourceService.cs
+++ b/src/LaunchDarkly.EventSource/EventSourceService.cs
@@ -94,7 +94,7 @@ namespace LaunchDarkly.EventSource
                 cancellationToken).ConfigureAwait(false))
             {
                 _logger.Debug("Response status: {0}", (int)response.StatusCode);
-                HandleInvalidResponses(response);
+                ValidateResponse(response);
 
                 OnConnectionOpened();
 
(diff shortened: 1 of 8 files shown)


src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -228,46 +228,23 @@ private HttpRequestMessage CreateHttpRequestMessage(Uri uri, string lastEventId)
...
+        private void ValidateResponse(HttpResponseMessage response)
         {
-            HandleUnsuccessfulStatusCodes(response);
-
-            // According to Specs, a client can be told to stop reconnecting using the HTTP 204 No Content response code
-            HandleNoContent(response);
-
-            // According to Specs, HTTP 200 OK responses that have a Content-Type specifying an unsupported type, 
-            // or that have no Content-Type at all, must cause the user agent to fail the connection.
-            HandleIncorrectMediaType(response);
```
eli-darkly, line 240: I cleaned up this logic a bit just to eliminate some extra helper functions that weren't really necessary, and to make it easier to see all the rules in one place.

src/LaunchDarkly.EventSource/EventSourceService.cs
```diff
@@ -228,46 +228,23 @@ private HttpRequestMessage CreateHttpRequestMessage(Uri uri, string lastEventId)
...
-        }
-
-        private void HandleNoContent(HttpResponseMessage response)
-        {
-            if (response.StatusCode == System.Net.HttpStatusCode.NoContent)
+            // Any non-2xx response status is an error. A 204 (no content) is also an error.
+            if (!response.IsSuccessStatusCode || response.StatusCode == System.Net.HttpStatusCode.NoContent)
             {
-                throw new EventSourceServiceUnsuccessfulResponseException(Resources.EventSource_204_Response, (int)response.StatusCode);
+                throw new EventSourceServiceUnsuccessfulResponseException((int)response.StatusCode);
```
eli-darkly, line 236: There's no longer a separate error message for complaining about a 204 response; it's just another unexpected status code. There can be a special meaning for 204 in SSE-based applications (the convention is that it means "don't bother trying to reconnect, there won't ever be any content here") but it's up to the caller to make decisions about that, which they can already do by looking at the StatusCode property of the exception.

src/LaunchDarkly.EventSource/Resources.resx
```diff
@@ -117,31 +117,19 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="EventSourceService_Read_Timeout" xml:space="preserve">
```
eli-darkly, line 120: I removed the `EventSourceService` from these names because all of these things already by definition apply to stuff in the `LaunchDarkly.EventSource` namespace.

src/LaunchDarkly.EventSource/Resources.resx
```diff
@@ -117,31 +117,19 @@
...
-    <value>A timeout occurred while reading the request from the remote EventSource API</value>
+  <data name="ErrorReadTimeout" xml:space="preserve">
+    <value>Read timeout elapsed with no new data from server; connection may have been silently dropped</value>
```
eli-darkly, line 121: I added the note about how it might mean the connection was silently dropped because that might save people a little time in troubleshooting— that is, the main point of using a read timeout (and heartbeats) for streaming connections is that connections can be silently dropped. Someone who doesn't know that might assume that the timeout error must indicate a server problem.

src/LaunchDarkly.EventSource/Resources.resx
```diff
@@ -117,31 +117,19 @@
...
-  </data>
-  <data name="EventSource_Logger_Closed" xml:space="preserve">
-    <value>EventSource.Close called</value>
-  </data>
-  <data name="EventSource_Logger_Connection_Error" xml:space="preserve">
-    <value>Encountered exception in EventSourceService.ConnectToEventSourceApi method. Exception Message: {0}</value>
-  </data>
-  <data name="EventSource_Logger_Disconnected" xml:space="preserve">
-    <value>EventSource Disconnected. Automatically delaying {0}ms before reconnecting.</value>
-  </data>
```
eli-darkly, line 143: These "Logger" ones weren't being used at all.

gwhelanLD (approved):

PR #73, opened 2021-05-03
use LaunchDarkly.TestHelpers for end-to-end HTTP testing

eli-darkly:
This updates the unit tests to use our [`dotnet-testhelpers`](https://github.com/launchdarkly/dotnet-testhelpers) project, instead of the combination of `EmbedIO` and fake HTTP message handlers that we were previously using. Advantages include:

* The library has the same target framework compatibility as EventSource, whereas EmbedIO didn't support .NET Framework 4.5.2. So we can now run the tests in all of our lowest compatible platform versions.
* It fully supports all of the HTTP behaviors we need to test and is pretty simple to use, so we can now run all of our HTTP tests as real end-to-end tests against a server, instead of relying on fake HttpMessageHandler implementations that don't actually do I/O.
* We maintain this library and it has no dependencies other than standard .NET APIs, so we shouldn't be blocked in the future by waiting for fixes or enhancements.


test/LaunchDarkly.EventSource.Tests/EventSourceEncodingTest.cs
```diff
@@ -52,26 +44,34 @@ private static string MakeLongString(int startNum, int endNum)
...
+            {
+                ret = ret.Then(Handlers.WriteChunkString(chunk, encoding));
+            }
+            return ret.Then(Handlers.Hang());
+        }
+
         [Theory]
         [InlineData(true)]
         [InlineData(false)]
         public void CanReceiveUtf8EventDataAsStrings(bool setExplicitEncoding)
```
eli-darkly, line 60: Most of the changes in this file are pretty straightforward and follow the same pattern as this test: instead of using a `StubMessageHandler` (which was a way of faking an HTTP client transaction without doing any I/O), it uses an `HttpHelpers.HttpServer` configured with some combination of `HttpHelpers` handlers. The changes look bigger than they are due to different indentation of the main block in each method. See PR comments for anything else worth noting.

test/LaunchDarkly.EventSource.Tests/EventSourceEndToEndTest.cs
```diff
@@ -1,184 +0,0 @@
...
-using Xunit.Abstractions;
-
-namespace LaunchDarkly.EventSource.Tests
-{
-    // For most of the EventSource tests, it's adequate to use a fake HttpMessageHandler that provides
-    // the desired response data without really doing HTTP. But we do want to be sure that our I/O logic
-    // works as intended with a real HTTP client, so this class does some basic EventSource tests
-    // against an embedded HTTP server.
-
-    public class EventSourceEndToEndTest : BaseTest
```
eli-darkly, line 21: I moved these tests into `EventSourceHttpBehaviorTest` because they are _all_ end-to-end tests now.

test/LaunchDarkly.EventSource.Tests/EventSourceHttpBehaviorTest.cs
```diff
@@ -17,153 +17,244 @@ public class EventSourceHttpBehaviorTest : BaseTest
...
+                    Assert.True(headers.All(
+                        item => req.Headers.Get(item.Key) == item.Value
+                    ));
+                }
+            }
         }
 
         [Fact]
-        public void ErrorForIncorrectContentType()
+        public void ReceiveEventStreamInChunks()
```
eli-darkly, line 148: Moved this and the next couple of tests from `EventSourceEndToEndTest`.

eli-darkly:
https://github.com/launchdarkly/dotnet-eventsource/pull/74 was accidentally merged into this branch, so the changes related to `ConnectionTimeout` and `ResponseStartTimeout` are from that PR and were already reviewed.

gwhelanLD (approved):

PR #75, opened 2021-05-06
use DocFX to generate HTML docs

eli-darkly:
This is a change to our release procedures that I'd like to integrate into our regular release tool for all of our .NET projects, and it'd be nice to use it for the upcoming LD .NET SDK 6.0 release, but it's easiest to test it on a single relatively independent project first.

We've been using Microsoft's obsolete Sandcastle Help File Builder tool to generate HTML documentation from doc comments. Besides being very old and unmaintained, it can only run in Windows, but for many years there was no real alternative (Doxygen is supposed to be compatible with C#, but I could never get it to work). However, in the last few years Microsoft has been improving its newer DocFX tool to the point where it seems usable for our purposes. DocFX has some idiosyncracies but it's maintained, and since it's a .NET Core app it can _theoretically_ run on Linux, which would be considerably more convenient for us.

The output of the tool looks pretty good - you can view an example [here](https://483-99271538-gh.circle-artifacts.com/0/artifacts/docs/index.html). The home page is very minimal, it would probably be better to just drop people directly into the API docs for the main mainspace but I haven't figured out if there's a way to do that.

Things I still need to figure out before we can use this more widely:

* So far I've been unable to run it on Linux - see [this issue](https://github.com/dotnet/docfx/issues/7347#issuecomment-833138151). The fact that they've tagged that issue as "v3" makes me think it might not get fixed in the current major version, so we may need to keep running it on a Windows CI host for releases. That's OK for now since we're using Windows for the rest of the release anyway, and it doesn't require nearly as much PowerShell scripting as the previous Sandcastle solution, but Linux would be simpler, especially now that .NET 5.0 will allow us to build all our code (even for .NET Framework) in a Linux container.
* I'm still a bit confused about how to integrate non-code-generated content into the docs. Specifically, C# doesn't have a standard way to document a _namespace_ (Sandcastle had a hack for this), and for our larger projects like the server-side .NET SDK it's desirable to have namespace summary pages that have some explanatory text, not just a list of classes. DocFX does have a way to add in Markdown files, but their documentation leaves something to be desired.

### DIFF
--- a/.ldrelease/zip-helper.ps1
+++ b/.ldrelease/zip-helper.ps1
@@ -1,0 +1,21 @@
+
+# see: https://stackoverflow.com/questions/27289115/system-io-compression-zipfile-net-4-5-output-zip-in-not-suitable-for-linux-mac
+class FixZipFilePathEncoding : System.Text.UTF8Encoding {
+    FixZipFilePathEncoding() : base($true) { }
+    [byte[]] GetBytes([string] $s)
+    {
+        $s = $s.Replace("\", "/");
+        return ([System.Text.UTF8Encoding]$this).GetBytes($s);
+    }
+}
+
+function Zip {
+    param(
+        [Parameter(Mandatory)][string]$sourcePath,
+        [Parameter(Mandatory)][string]$zipFile
+    )
+    [Reflection.Assembly]::LoadWithPartialName( "System.IO.Compression.FileSystem" ) | Out-Null
+    [System.IO.Compression.ZipFile]::CreateFromDirectory($sourcePath, $zipFile, `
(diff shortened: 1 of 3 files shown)


gwhelanLD (approved):

.ldrelease/zip-helper.ps1
```diff
@@ -0,0 +1,21 @@
+
```
eli-darkly, line 1: This is just a copy of the logic we needed from the built-in Releaser helper scripts. As long as this implementation is being done on a per-project basis rather than being part of our project templates, I figured it was best not to rely on details of those built-ins.

Issue #77, opened 2021-05-15
CancellationTokenSource memory leak


s2maki:
Hello,

In using your library to process SSE events, I'm finding that as millions of events are received there can be a significant buildup in RAM.  I ran it through dotMemory to profile and found that the leak appears to be due to the combined token in DoWithTimeout (AsyncHelpers.cs).  The combined token is used but never disposed.  Adding a `combinedCancellation.Dispose()` after the `await task` seems to cure the problem.

Thanks!

eli-darkly:
Since you mentioned AsyncHelpers.cs, I'm assuming that you're using version 4.0.0 or later, but can you confirm what version you're using?

eli-darkly:
Also— just to be clear, do you have reason to think this is a _leak_ in the sense of leaking resources that remain in use, or more of a case of "under heavy usage, these ephemeral objects build up before the garbage collector gets around to cleaning them up"? The latter seems more likely as I don't see how a reference to this `CancellationTokenSource` instance would be retained, but it's hard to be sure without any more details on what you observed.

eli-darkly:
I see no reason not to add an explicit dispose here (or a `using`) in any case, but just wanted to get a little more clarity on the issue.

s2maki:
I was originally using 4.0.0, but I did upgrade to 4.1.1 and was still able to reproduce the issue.  I further tested by downloading the source and building myself out of master.  It still reproduced.

It's a proper leak; the GC never cleans the tokens up.  I've seen my process get up to 2GB worth of combinedCancellation token objects over a course of an hour or two, all sitting in Generation 2 heap.  Once the dispose is added, the objects never even make it to Generation 1.

Thanks!

eli-darkly:
Thanks. Yes, we should definitely fix this and the fix is straightforward, so I should have a patch out shortly. I'm still not sure why CreatedLinkedTokenSource behaves this way (I see some mentions online of a similar finding, but no details on where the leaked reference is being retained or why) but the .NET API docs do say specifically that one ought to call Dispose, implying that Microsoft is aware of some reason why GC won't work as desired in this case.

eli-darkly:
[...]

PR #78, opened 2021-05-17
dispose of all CancellationTokenSource instances explicitly to avoid leaks

eli-darkly:
See https://github.com/launchdarkly/dotnet-eventsource/issues/77.

There's only one other place where we create a `CancellationTokenSource`: [in EventSource.StartAsync](https://github.com/launchdarkly/dotnet-eventsource/blob/4.1.1/src/LaunchDarkly.EventSource/EventSource.cs#L145). There, we're creating one for the HTTP request we're about to make, and we don't want to dispose of it there because the stream connection is meant to remain open. We dispose of it in [EventSource.CancelCurrentRequest](https://github.com/launchdarkly/dotnet-eventsource/blob/4.1.1/src/LaunchDarkly.EventSource/EventSource.cs#L277), which always gets called either if we need to drop the stream and start a new request, or if the whole EventSource is being disposed of. So we should not have to worry about a leak there.

There's no new test coverage here, because we don't AFAIK have a good way to see these kinds of heap usage details from within test code (without adding a whole new test framework such as the JetBrains suite). However, it was easy to verify the fix by running a piece of test code that simply called DoWithTimeout with a no-op task within a tight loop: with the fix heap usage remained stable, but without the fix it increased the longer I left the test running.

### DIFF
--- a/src/LaunchDarkly.EventSource/AsyncHelpers.cs
+++ b/src/LaunchDarkly.EventSource/AsyncHelpers.cs
@@ -20,31 +20,35 @@ namespace LaunchDarkly.EventSource
             Func<CancellationToken, Task<T>> taskFn
             )
         {
-            var timeoutCancellation = new CancellationTokenSource(timeout);
-            var combinedCancellation = CancellationTokenSource.CreateLinkedTokenSource(
-                cancellationToken, timeoutCancellation.Token);
-            Task<T> task = null;
-            try
+            using (var timeoutCancellation = new CancellationTokenSource(timeout))
             {
-                task = taskFn(combinedCancellation.Token);
-                var result = await task;
-                return result;
-            }
-            catch (AggregateException e) when (e.InnerException is OperationCanceledException)
-            {
-                if (cancellationToken.IsCancellationRequested)
+                using (var combinedCancellation = CancellationTokenSource.CreateLinkedTokenSource(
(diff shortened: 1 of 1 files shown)


src/LaunchDarkly.EventSource/AsyncHelpers.cs
```diff
@@ -20,31 +20,35 @@ internal static class AsyncHelpers
...
-                {
-                    throw e.InnerException;
-                }
-                throw new ReadTimeoutException();
-            }
-            catch (OperationCanceledException)
-            {
-                if (cancellationToken.IsCancellationRequested)
+                using (var combinedCancellation = CancellationTokenSource.CreateLinkedTokenSource(
+                    cancellationToken, timeoutCancellation.Token))
```
eli-darkly, line 26: The indentation change makes this diff hard to read, but the change here is simply that both `timeoutCancellation` and `combinedCancellation` are now scoped in `using` blocks so they'll be `Dispose`d in LIFO order no matter how we leave this method. Based on online discussions I've seen it seems that `CreateLinkedTokenSource` is more of a leak risk than `new CancellationTokenSource(TimeSpan)`, but there's no reason _not_ to dispose of them both of explicitly.

gwhelanLD (approved):

PR #83, opened 2021-11-13
WIP: implement contract tests

eli-darkly:
See: https://github.com/launchdarkly/sse-contract-tests/pull/1

An example of this running in CI is [here](https://app.circleci.com/pipelines/github/launchdarkly/dotnet-eventsource/221/workflows/600e77f3-7d0a-4210-95b9-baa9f3a6c2e2/jobs/615). That job is currently failing, as it should, since this SSE implementation is not quite compliant with spec; if we want CI to pass, we'll need to either disable the failing tests or fix the implementation. Specifically:

* It does not work with CR-only line terminators.
* In some cases, messages can be delivered out of order due to how it's using asynchronous tasks.

Those bugs do not affect use in LaunchDarkly, because our services send LF terminators and we already have a versioning mechanism to handle out-of-order messages, but we should still fix them.


contract-tests/README.md
```diff
@@ -0,0 +1,5 @@
+# SSE client contract test service
+
+This directory contains an implementation of the cross-platform SSE testing protocol defined by https://github.com/launchdarkly/sse-contract-testing. See that project's `README` for details of this protocol, and the kinds of SSE client capabilities that are relevant to the contract tests. This code should not need to be updated unless the SSE client has added or removed such capabilities.
```
kinyoklion, line 3: This is a draft, so commenting may be pre-mature, but I don't think this URL is correct.
 I see: https://github.com/launchdarkly/sse-contract-tests but not the URL specified in this comment (testing ->tests).

eli-darkly: Ugh, indeed

eli-darkly: Of course I pasted that same wrong content into a half-dozen repos.

PR #85, opened 2022-03-03
added onhttprequest event hook to enable blazor support

acquleo:
A OnHttpRequest event is raised before the request in order to be able to execute the SetBrowserResponseStreamingEnabled method on a Blazor app client.


eli-darkly (changes requested):
Providing a hook for general modification of request attributes is a reasonable feature. However, I think the `event` pattern is not a great fit for this. All of the other events defined on `EventSource` are for _notifications_ of things that have happened— not for making changes to the behavior of `EventSource`. And an `event` inherently can have any number of handlers, which isn't a good fit for this feature since the ordering of event handler calls is not well defined but the ordering of changes to an HTTP request could be significant.

So, instead of exposing an `event` for this, I would rather provide a `ConfigurationBuilder` setter: `RequestModifier(Action<HttpRequestMessage>)`. If this hook is non-null, it is called by `EventSourceService` at the same place that you are currently calling `OnHttpRequest?.Invoke`.

acquleo:
I've modified the code according to your proposal.

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -190,6 +196,9 @@ internal Configuration(ConfigurationBuilder builder)
             RequestHeaders = new Dictionary<string, string>(builder._requestHeaders);
             ResponseStartTimeout = builder._responseStartTimeout;
             RequestBodyFactory = builder._requestBodyFactory;
+            HttpRequestModifier = builder._httpRequestModifier; 
+
+
```
eli-darkly, line 201: Minor: extra blank lines here.

src/LaunchDarkly.EventSource/Configuration.cs
```diff
@@ -93,6 +93,12 @@ public sealed class Configuration
         /// <seealso cref="ConfigurationBuilder.HttpMessageHandler(HttpMessageHandler)"/>
         public HttpMessageHandler HttpMessageHandler { get; }
 
+        /// <summary>
+        /// Delegate hook invoked before an http request has been performed.
```
eli-darkly, line 97:
```suggestion
        /// Delegate hook invoked before an HTTP request has been performed.
```

src/LaunchDarkly.EventSource/ConfigurationBuilder.cs
```diff
@@ -77,6 +77,18 @@ internal ConfigurationBuilder(Uri uri)
         public ConfigurationBuilder ConnectionTimeout(TimeSpan responseStartTimeout) =>
             ResponseStartTimeout(responseStartTimeout);
 
+
+        /// <summary>
+        /// Sets a delegate hook invoked before an http request has been performed.
```
eli-darkly, line 82:
```suggestion
        /// Sets a delegate hook invoked before an HTTP request has been performed, allowing changes to the request.
```

src/LaunchDarkly.EventSource/EventSource.cs
```diff
@@ -302,6 +302,7 @@ private async Task ConnectToEventSourceAsync(CancellationToken cancellationToken
 
             var svc = GetEventSourceService(_configuration);
 
+
```
eli-darkly, line 305: Extra blank line.

src/LaunchDarkly.EventSource/LaunchDarkly.EventSource.csproj
```diff
@@ -49,5 +49,6 @@
 
   <PropertyGroup>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\LaunchDarkly.EventSource.xml</DocumentationFile>
+    <AssemblyOriginatorKeyFile>../../LaunchDarkly.EventSource.snk</AssemblyOriginatorKeyFile>
```
eli-darkly, line 52: I don't know why this change was made, but please revert it. We do not want the assembly to be strong-named in _every_ configuration, but only in the Release configuration, as is specified on line 33 of this file. Possibly your IDE made this change automatically for some reason, but please be careful that you're not checking in unintended changes to the project configuration.

eli-darkly (changes requested):

acquleo (commented):
agree with your comments

eli-darkly:
This feature is now in the 4.2.0 release. Sorry it took so long for us to get back to it.

PR #93, opened 2023-01-18
LaunchDarkly.EventSource 6.0 rewrite, part 1

eli-darkly:
This is a major rethink of this fairly old code, closely based on the [v4.0.0 redesign](https://github.com/launchdarkly/okhttp-eventsource/releases/tag/4.0.0) of our Java SSE client `okhttp-eventsource`. It has the following goals:

* Give the caller more control over when and where things are executed, by reversing the overall design from a push model (SSE client runs a worker task which calls event handlers) to a pull model (caller explicitly requests events one at a time, but _can_ run a separate worker task for this if they want to).
* Properly support asynchronous event handling. The previous version used async I/O, but when it called event handlers, they were regular synchronous void functions.
* Make it more feasible to implement a "read data for a large event incrementally" mode, which what `okhttp-eventsource` calls [`streamEventData`](https://javadoc.io/static/com.launchdarkly/okhttp-eventsource/4.1.0/com/launchdarkly/eventsource/EventSource.Builder.html#streamEventData(boolean)). This PR does not implement that feature, but it's an important first step toward it, because (as I found when trying to implement it in an earlier version of `okhttp-eventsource`) the feature doesn't play well with the choice of a push model.
* Improve separation of concerns in a way that makes the implementation cleaner, more flexible, and easier to test. The retry delay logic now uses a `RetryDelayStrategy` abstraction, where the default implementation is our usual backoff/jitter; the logic for obtaining an input stream now uses a `ConnectStrategy` abstraction, where the default implementation is HTTP. There's also an `ErrorStrategy` abstraction which is fairly similar to the previous `ConnectionErrorHandler`, but its behavior is now a bit better defined.
* Fix areas of noncompliance with the SSE spec— primarily that we weren't correctly handling CR-only line endings, which was a side effect of relying on .NET's StreamReader class.

Like the Java equivalent, this PR provides a `BackgroundEventSource` wrapper class that is the equivalent of the old push model, making it relatively straightforward to adapt any code that was using the older version; the main difference is that the event handlers must return `Task` (which doesn't mean they have to really be async— synchronous logic can be kept by adding `return Task.CompletedTask`).


src/LaunchDarkly.EventSource/Internal/BufferedLineParser.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// To use this class, we repeatedly call <see cref="ReadAsync"/> to obtain a piece of data.
+    /// Rather than copying this back into a buffer provided by the caller, BufferedLineParser
+    /// exposes its own fixed-size buffer directly and marks the portion being read; the caller
+    /// is responsible for inspecting this data before the next call to <see cref="ReadAsync"/>.
+    /// </para>
+    /// <para>
+    /// This class is not thread-safe.
+    /// </para>
+    /// </remarks>
+    internal class BufferedLineParser
```
eli-darkly, line 23: The logic here is a pretty straight port from [the equivalent class](https://github.com/launchdarkly/okhttp-eventsource/blob/main/src/main/java/com/launchdarkly/eventsource/BufferedLineParser.java) in `okhttp-eventsource`. The differences are basically 1. it uses async I/O and 2. it abstracts the stream with a read function rather than referencing the Stream class (see code comment immediately below).

src/LaunchDarkly.EventSource/Internal/EventParser.cs
```diff
@@ -0,0 +1,307 @@
...
+using LaunchDarkly.Logging;
+
+using static LaunchDarkly.EventSource.Internal.AsyncHelpers;
+
+namespace LaunchDarkly.EventSource.Internal
+{
+    /// <summary>
+    /// An internal class containing helper methods to parse Server Sent Event data.
+    /// </summary>
+    internal sealed class EventParser
```
eli-darkly, line 16: Again this is a pretty direct port of [the Java code](https://github.com/launchdarkly/okhttp-eventsource/blob/main/src/main/java/com/launchdarkly/eventsource/EventParser.java), but it is smaller because it doesn't yet include any of the "provide a stream for the caller to read event data incrementally" logic.

src/LaunchDarkly.EventSource/Resources.Designer.cs
```diff
@@ -94,20 +94,29 @@ internal class Resources {
...
+                return ResourceManager.GetString("StreamClosedByCaller", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to the stream was closed from the server side.
+        /// </summary>
+        internal static string StreamClosedByServer {
+            get {
+                return ResourceManager.GetString("StreamClosedByServer", resourceCulture);
```
eli-darkly, line 119: I'm not sure how important it is to provide these internationalizable strings— given that 1. we're still only providing English-language text, and 2. it's not something we're doing in any of our other libraries— but since this project has historically had them, I've maintained them.

src/LaunchDarkly.EventSource/RetryDelayStrategy.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <para>
+    /// Subclasses should be immutable. To implement strategies where the delay uses
+    /// different parameters on each subsequent retry (such as exponential backoff),
+    /// the strategy should return a new instance of its own class in
+    /// <see cref="Result.Next"/>, rather than modifying the state of the existing
+    /// instance. This makes it easy for EventSource to reset to the original delay
+    /// state when appropriate by simply reusing the original instance.
+    /// </para>
+    /// </remarks>
+    public abstract class RetryDelayStrategy
```
eli-darkly, line 25: Equivalent to the [Java version](https://github.com/launchdarkly/okhttp-eventsource/blob/main/src/main/java/com/launchdarkly/eventsource/RetryDelayStrategy.java).

src/LaunchDarkly.EventSource/ErrorStrategy.cs
```diff
@@ -0,0 +1,152 @@
...
+{
+    /// <summary>
+    /// An abstraction of how to determine whether a stream failure should be thrown to the
+	/// caller as an exception, or treated as an event.
+    /// </summary>
+    /// <para>
+    /// Instances of this class should be immutable.
+    /// </para>
+	/// <seealso cref="ConfigurationBuilder.ErrorStrategy(ErrorStrategy)"/>
+    public abstract class ErrorStrategy
```
eli-darkly, line 16: Equivalent to the [Java version](https://github.com/launchdarkly/okhttp-eventsource/blob/main/src/main/java/com/launchdarkly/eventsource/ErrorStrategy.java).

src/LaunchDarkly.EventSource/ConnectStrategy.cs
```diff
@@ -0,0 +1,163 @@
...
+    /// Or, if you want to consume an input stream from some other source, you can
+    /// create your own subclass of ConnectStrategy.
+    /// </para>
+    /// <para>
+    /// Instances of this class should be immutable and not contain any state that
+    /// is specific to one active stream.The {@link ConnectStrategy.Client}
+    /// that they produce is stateful and belongs to a single EventSource.
+    /// </para>
+    /// </remarks>
+    public abstract class ConnectStrategy
```
eli-darkly, line 34: Equivalent to the [Java version](https://github.com/launchdarkly/okhttp-eventsource/blob/main/src/main/java/com/launchdarkly/eventsource/ConnectStrategy.java).

src/LaunchDarkly.EventSource/IEventSource.cs
```diff
@@ -54,25 +86,90 @@ public interface IEventSource
...
+        /// </para>
+        /// <para>
+        /// This method must be called from the same thread that first started using the
+        /// stream (that is, the thread that called <see cref="StartAsync"/> or read the
+        /// first event).
+        /// </para>
+        /// </remarks>
+        /// <returns>an SSE message</returns>
+        /// <seealso cref="ReadAnyEventAsync"/>
+        Task<MessageEvent> ReadMessageAsync();
```
eli-darkly, line 118: One notable API difference between this and the Java `okhttp-eventsource` is that the latter also has methods that return messages/events as an `Iterable` (Java equivalent of an `IEnumerable`), so you can just treat them like any collection instead of explicitly asking for each event. That's convenient, but the problem is that _async_ generators are a C# 8.0 feature, and we're still constrained to use C# 7.3 in order to maintain .NET Standard 2.0 compatibility.

PR #94, opened 2023-01-19
add streaming event data mode (6.0 rewrite, part 2)

eli-darkly:
This adds the incremental loading feature that was one of the main motivations for the whole rewrite. As before, the implementation is closely based on the Java version in `okhttp-eventsource`. Briefly:

* By default, everything works the same as before except that there is no longer any "store the message data as a reference to an internal byte buffer" behavior— it just stores the data as a string. The intention is that if an application is going to be reading large enough data that the string type would be inefficient, then they will use this new mode instead.
* If you set `StreamEventData(true)`, then when you receive a `MessageEvent` it contains a special stream, accessed via the `DataStream` property. When you read from this stream, it is really reading from the `EventParser`, consuming the value of every `data:` field up till it hits a blank line, at which point that stream returns EOF.
* As soon as you request another event, the `DataStream` for the previous event becomes invalid, and `EventParser` skips over any content that hasn't yet been consumed from that one.

One wrinkle with this, which is a necessary departure from the SSE spec, is that `EventParser` has to return the `MessageEvent` before it has seen anything past the first `data:` line. So at that point, it doesn't yet know for sure that there really will be a full event before the end of the stream— and, if any other fields like `event:` or `id:` appear _after_ `data:`, it can't put them into the `MessageEvent`. This is not an issue for using EventSource in the LaunchDarkly SDKs, because 1. until all of the data is consumed we won't actually be storing any of it, just parsing it; 2. the LD services always send `event:` before `data:`; and 3. with the new `ExpectFields` option, the SDK can specify that it definitely needs to see an `event:` field and so if for some reason the fields were sent out of order, EventSource will just go back to buffering the full event like before.


src/LaunchDarkly.EventSource/Events/Utf8ByteSpan.cs
```diff
@@ -1,128 +0,0 @@
...
-    /// <summary>
-    /// Points to a span of UTF-8-encoded text in a buffer.
-    /// </summary>
-    /// <remarks>
-    /// This is similar to the <c>Span</c> type in .NET 5. It is used internally by
-    /// <c>EventSource</c> to store event data if the stream is using UTF-8 encoding.
-    /// If so, reading <see cref="MessageEvent.DataUtf8Bytes"/> will return the same
-    /// byte array, to avoid unnecessary copying.
-    /// </remarks>
-    public struct Utf8ByteSpan
```
eli-darkly, line 15: This is no longer a public type because `MessageEvent` no longer has a `DataUtf8Bytes` property. It's now in `Internal` and is used only as a convenience container for a tuple of (buffer, offset, length).

src/LaunchDarkly.EventSource/Internal/AsyncHelpers.cs
```diff
@@ -6,6 +6,23 @@ namespace LaunchDarkly.EventSource.Internal
...
+        // In general, mixing sync and async code is not recommended, and if done in other ways can result in
+        // deadlocks. See: https://stackoverflow.com/questions/9343594/how-to-call-asynchronous-method-from-synchronous-method-in-c
+        // Task.Wait would only be safe if we could guarantee that every intermediate Task within the async
+        // code had been modified with ConfigureAwait(false), but that is very error-prone.
+
+        internal static T WaitSafely<T>(Func<Task<T>> taskFn) =>
+            _taskFactory.StartNew(taskFn)
+                .Unwrap()
+                .GetAwaiter()
+                .GetResult();
```
eli-darkly, line 23: This is used only to implement the synchronous `Read` method in the special stream provided by EventParser.

PR #99, opened 2023-10-04
feat: Update minimum version of LaunchDarkly.Logging to 2.0.0

kinyoklion:


src/LaunchDarkly.EventSource/LaunchDarkly.EventSource.csproj
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="LaunchDarkly.Logging" Version="[1.0.1,3.0.0)" />
+    <PackageReference Include="LaunchDarkly.Logging" Version="[2.0.0,3.0.0)" />
```
kinyoklion, line 21: Because .Net will install the oldest compatible version it is most important to have the minimum be the same for all of then. This will help people who have problems with the binding redirects.

tanderson-ld (approved):

PR #100, opened 2023-10-05
chore: Update codeowners to dotnet team.

kinyoklion:


CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Repository Maintainers
+* @launchdarkly/team-sdk-net
```
kinyoklion, line 2: Looks like I need to do some other work first.

louis-launchdarkly (approved):

PR #104, opened 2025-02-21
feat: Report event source headers on open.

kinyoklion:


src/LaunchDarkly.EventSource/StateChangedEventArgs.cs
```diff
@@ -16,13 +17,23 @@ public class StateChangedEventArgs : EventArgs
         /// </value>
         public ReadyState ReadyState { get; }
 
+        /// <summary>
+        /// Get the response headers. Only populated for <see cref="ReadyState.Open"/>.
+        /// <value>
+        /// A collection of header values when the ReadyState is Open, or null.
+        /// </value>
+        /// </summary>
+        public IEnumerable<KeyValuePair<string,IEnumerable<string>>> Headers { get; }
```
kinyoklion, line 26: This is the public facing change.

src/LaunchDarkly.EventSource/StateChangedEventArgs.cs
```diff
@@ -16,13 +17,23 @@ public class StateChangedEventArgs : EventArgs
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="StateChangedEventArgs"/> class.
         /// </summary>
         /// <param name="readyState">One of the <see cref="EventSource.ReadyState"/> values, which represents the state of the EventSource connection.</param>
-        public StateChangedEventArgs(ReadyState readyState)
+        /// <param name="headers">Response headers when the <see cref="StateChangedEventArgs.ReadyState"/> is <see cref="ReadyState.Open"/>. Otherwise null.</param>
+        public StateChangedEventArgs(ReadyState readyState, IEnumerable<KeyValuePair<string,IEnumerable<string>>> headers = null)
```
kinyoklion, line 33: This constructor is public, but isn't something someone would really need to use in normal operation.

Likely it should have been internal.

tanderson-ld (approved):

PR #126, opened 2026-08-31
feat!: temporary retry-delay bounds and SSE retry: field handling

tanderson-ld:
## Summary

Gives consumers runtime control over reconnection timing, applies the SSE `retry:` field to the backoff instead of discarding it, and fixes several defects in the backoff path.

**New API** — breaking, since `IEventSource` gains members. Implementers must add them; callers of `EventSource` are unaffected.

```csharp
void SetTemporaryRetryDelayBounds(TimeSpan initialDelay, TimeSpan maxDelay)
void ClearTemporaryRetryDelayBounds()
```

Temporary bounds revert automatically once a connection stays open for `BackoffResetThreshold`, or explicitly via `Clear`. Reapplying identical bounds is idempotent, so a caller may set them on every failure without pinning the delay at the minimum.

**`retry:` field.** A server-directed value replaces the minimum delay, is clamped to 1 hour, resets the backoff formula input so the next attempt uses the directed value directly, and is sticky until the server sends another. Ceilings and the active regime are never affected.

Note the 1-hour cap is unobservable at default configuration: the configured `MaxRetryDelay` of 30s bounds the delay first, so a directed value above 30s is reduced to it. This matches Java, whose `DefaultRetryDelayStrategy` applies the same clamp with the same 30s default. Go honors larger values only because it ships no default ceiling at all.

## Fixes

- The delay computation raised 2 to an unbounded exponent in floating point and clamped the result, throwing `OverflowException` past exponent 21 when the maximum exceeded `int.MaxValue` ms. That exception escaped the reconnect loop, stopping the stream permanently. Now integer-only, comparing against the ceiling before shifting.
- A pending backoff wait could not be interrupted by `Close()`. Previously bounded by `MaxRetryDelay` (30s default); this change makes an hour reachable, so it matters more. RETRY §1.10.1.
- `InitialRetryDelay(TimeSpan.Zero)` disabled backoff outright, because the parsed `retry:` value doubled as an on/off gate. That gate is gone.
- A computed delay above `Task.Delay`'s platform ceiling threw out of `StartAsync`, stopping the stream permanently. Clamped at `int.MaxValue` ms, the lower of the two target frameworks' ceilings.
- A large `retry:` value threw `OverflowException` from `TimeSpan.FromMilliseconds` before the 1-hour clamp could apply, tearing the connection down on every read of the line and applying nothing. Now clamped in the integer domain at the wire boundary, which is where Java and Go clamp. The overflow is symmetric; both directions are covered.
- The healthy-operation reset window was measured with `DateTime.Now`, a *local* wall clock, so a daylight-saving transition was a deterministic twice-yearly trigger and an NTP step could fire the reset early or suppress it entirely. Now measured with a monotonic `Stopwatch`.

## Observable behavior changes

- `ReadyState` after `Close()` on a never-started source goes `Raw` → `Shutdown` rather than staying `Raw`, and `Closed(Shutdown)` now fires on that path where it was previously silent.
- A server-directed `retry:` value now determines the delay *duration*. Previously the backoff was constructed once from `InitialRetryDelay` and the `retry:` value only gated whether any wait happened, so every positive value behaved identically and zero disabled the wait entirely.
- Reconnect delays above `int.MaxValue` ms (~24.9 days) are clamped rather than escaping `StartAsync`.
- The healthy-operation window is measured monotonically and is unaffected by clock adjustments.
- A sub-millisecond `InitialRetryDelay` no longer grows into a normal backoff progression. The floating-point path kept such a value alive until doubling lifted it past 1ms; the integer path truncates it to zero up front.

## Worth a reviewer's attention

- **`n` resets on a `retry:` hint**, matching Java and Go, which diverges from RETRY §1.11.2 ("MUST NOT alter the `attempts` counter"). All three implementations do this; worth a spec question rather than three separate deviations.
- **`max == 0` means zero delay**, not "no ceiling, use the base" as in Java and Go. Pre-existing .NET behavior, preserved by the parity vectors, and the literal reading of an API that always passes both bounds explicitly. Java and Go treat 0 as an *unset* sentinel, which this API does not have.
- **A bounds change resets the backoff progression.** Java's `activateRetryDelayStrategy` is a documented pointer swap that never resets, with per-strategy state. Here the reset is load-bearing: raising the minimum while `n` is large would compute `min << n` and pin at the ceiling immediately. Consequence for the consumer: the extended regime must raise the minimum, not just the ceiling.
- `EventSource` gains an `internal BackOff` accessor used only by tests.
- Implemented with the backoff owning the active bounds rather than layering a nullable override on `EventSource`; one less piece of state.

## Testing

296 unit tests, including:

- 85 parity vectors generated from the replaced algorithm, so the rewrite is pinned to the delays the library produced before it.
- An exact `BigInteger` reference for the delay formula over curated bound pairs and a million randomized checks, pinning the shift-based arithmetic to a reference that cannot overflow. Removing the `n >= 63` shift guard fails it with `min=1 max=2 n=64 actual=1 expected=2` — C# masking `<< 64` into `<< 0`.
- An end-to-end test that a `retry:` value changes the *actual wait*, not just the backoff's internal state. Asserts a lower bound only, since jitter makes `T/2` a deterministic floor.

SSE contract tests pass with no new skips. Fixes were mutation-verified — the implementation was temporarily broken in each case to confirm the tests fail. `net462` unit tests run only on Windows, so CI covers that leg.

SDK-2791
SDK-2855

---

> [!NOTE]
> **Overview**
> Adds **runtime temporary retry bounds** on `IEventSource` (`SetTemporaryRetryDelayBounds` / `ClearTemporaryRetryDelayBounds`), auto-cleared after a connection stays open for `BackoffResetThreshold` (or explicitly). Reaplying the same bounds is idempotent so error handlers can set them on every failure without pinning delay at the minimum.
> 
> **SSE `retry:`** now drives reconnect timing via a sticky server-directed minimum (clamped to 1 hour at parse time, including huge values that would overflow `TimeSpan`), resets backoff `n`, and respects configured/temporary ceilings without reverting temporary max.
> 
> **Backoff rewrite** (`ExponentialBackoffWithDecorrelation`): integer shift-based delay math with overflow guards, thread-safe state, and fixes where overflow or oversized delays could fault `StartAsync` permanently. **Reconnect loop**: backoff waits use a shutdown token so `Close()` can interrupt long waits; healthy-window reset uses `Stopwatch` instead of `DateTime.Now`. **`Close()`** on a never-started source transitions to `Shutdown` and raises `Closed(Shutdown)`; fixes tight reconnect spin when closed before `StartAsync`.
> 
> Breaking change: `IEventSource` gains two methods (implementers must add them).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d02357469669472b2e13edd8067a04611bb4328. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
```diff
@@ -2,47 +2,205 @@
...
-            _reconnectAttempts++;
-            return TimeSpan.FromMilliseconds(nextDelay);
+            lock (_lock)
+            {
+                long unjittered = GetUnjitteredMillisecondsForN(_backoffN);
+
+                // Random.Next takes an int bound. 2^31 milliseconds is far longer than any
+                // reconnect delay we would use, so saturating here cannot affect a realistic
+                // delay.
+                int jitterBound = unjittered > int.MaxValue ? int.MaxValue : (int)unjittered;
```
jsonbailey: nit: we seem to do the math.min / max for value clamping. Any reason we don't do it here?
```suggestion
                int jitterBound = Math.min(unjittered, int.MaxValue);
```

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
```diff
@@ -2,47 +2,205 @@
...
+        /// n resets, so the next attempt uses the directed value itself rather than continuing to
+        /// double from wherever the progression had reached.
+        /// </para>
+        /// <para>
+        /// Negative values become zero. No upper bound is applied here; a caller taking a value
+        /// from an untrusted source is responsible for capping it first.
+        /// </para>
+        /// </remarks>
+        /// <param name="minDelay">the server-directed reconnection time</param>
+        public void SetServerDirectedMinDelay(TimeSpan minDelay)
```
jsonbailey, line 119: I think we need to discuss this. If we don't trust the server to always send us safe values, having this unguarded could cause serious problems. I also don't know that setting this should reset the backoffN. We do guard against a negative number but that sets the value to 0. So if the server sent us a value 0 or negative, it could slam us with requests and never wait. If it sent us a 0 by accident and then later had an error, the event source would hit without a wait due to calculating a delay increase based on 0.

If it continually sent us a 0 or negative, the backoff would never advance due to n being reset.

jsonbailey (changes requested):

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
tanderson-ld: This is existing behavior and six of eight implementations honor zero, and all eight honor any value greater than zero, including 1 millisecond.  This agrees with the spec https://html.spec.whatwg.org/multipage/server-sent-events.html .

The project's objective is to prevent the SDK from giving up forever, the least deviation from the SSE spec to achieve that is to clamp the upper side of the retry domain.

I don't disagree that clamping the lower side could be useful to us, but I argue it is not required to complete the project.

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
tanderson-ld: Good catch — that ternary was the odd one out. `EventSource` already clamps the server-directed `retry:` value with `Math.Max`/`Math.Min`, so this now matches:

```csharp
int jitterBound = (int)Math.Min(unjittered, int.MaxValue);
```

Two small adjustments from the suggestion as written: C# spells it `Math.Min`, and since `unjittered` is a `long` the result of `Math.Min` is also a `long`, so the assignment needs an explicit `(int)` cast.

Behaviorally identical either way — `GetUnjitteredMillisecondsForN` returns either `0` or a positive value and never a negative one, so the `Math.Min` result is always within `int` range and the narrowing cast cannot lose information. Full suite still passes (296/296).

Landing in an incoming commit.

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
jsonbailey: I am fine with this as is, but we should at least create a task to track this for future work.

tanderson-ld: Work item created.

jsonbailey (approved):

src/LaunchDarkly.EventSource/ExponentialBackoffWithDecorrelation.cs
tanderson-ld: Worth the discussion — I surveyed all eight LaunchDarkly eventsource implementations before answering, and your fourth point turns out to be the sharpest of the five.

**On the arithmetic.** You're right that a `0` poisons the progression, and it's worse than "no wait on the next attempt": with a base of zero, `0 × 2^n = 0` for every `n`. The backoff isn't slowed, it's mathematically dead, and it stays dead until a later `retry:` replaces the base. That's the real defect shape, not the immediate reconnect.

That also reframes what Rust's fix is for. `rust-eventsource-client` floors a server value at **1 ms** with the comment *"A server that sends `retry: 0` would otherwise collapse the backoff to zero and reconnect would become a tight loop."* 1 ms doesn't rate-limit anything — its purpose is keeping `1ms × 2^n` able to climb to the ceiling.

**What the eight implementations do today:**

| | `retry: 0` | negative |
|---|---|---|
| rust-eventsource-client | floored at 1 ms | — |
| eventsource (Go) | ignored — `if pub.Retry() > 0` | ignored |
| okhttp-eventsource | honored | rejected at parse (`^[\d]+$`) |
| swift-eventsource | honored | rejected at parse |
| ruby-eventsource | honored (explicit `if target == 0` branch) | rejected at parse |
| python-eventsource | honored | accepted |
| js-eventsource | honored | accepted |
| dotnet-eventsource | honored | accepted |

Six of eight honor it. Ruby even has a deliberate zero branch, so it was considered and allowed there.

**Why this isn't a regression in this PR.** `main` already honored `retry: 0` as "no wait at all" — `MaybeWaitWithBackOff` was gated on `if (_retryDelay.TotalMilliseconds > 0)`, so a zero skipped the wait *and* the log line entirely, and never advanced the attempt count. Same observable outcome as this branch. What this PR adds on that path is the 1-hour clamp, which `main` didn't have.

**On guarding it.** I'd like to, but the spec currently forbids both available fixes. RETRY 1.11.4 says *"No floor is imposed; server-directed values at or below the cap are applied verbatim"* — which makes Rust's floor and Go's ignore-non-positive both non-conformant. So the two implementations that are safe are safe by deviating, and picking either here would add a third behavior to an already-split fleet. I think that language needs revisiting (it reads as though it was written to stop implementations overriding a legitimately small value like a `429` with `Retry-After: 10`, not to require honoring zero), and that's a cross-SDK conversation rather than something to settle in this PR.

**On resetting the counter — you're right, and against three other SDKs.** RETRY 1.11.2 is explicit: a server-directed wait duration **MUST NOT** alter the `attempts` counter. Java's `resetAllRegisteredStrategyState()`, Go's `retryCount = 0` per profile, and JS's `setBaseDelay` all do it; Swift and Ruby don't. So four of eight are non-conformant on that requirement, .NET included. Also a cross-SDK item rather than a unilateral change, but it should be on the list.

One more thing worth recording: the "a later `retry:` can fix it" mitigation doesn't hold where it matters. Receiving a `retry:` line requires establishing a connection and reading the body — so during the failure window where backoff actually matters, the corrective value can't arrive. Once connections succeed again the loop has already stopped. It's latent-armed rather than self-healing.

So: no change here, and two items for the cross-SDK spec discussion — the zero floor and the counter reset.

